PR #437, opened 2025-11-25
Pretty print using Vite logger

### LINKED ISSUE #206: Vite plugin: Pretty print via Vite's logger
For the vite plugin, we can use Vite's logger to Pretty print (uses picocolors internally)

example code:

```ts
// vite-plugin-validate-env.ts
import type { Plugin } from 'vite'
import { loadEnv } from 'vite'
import { validateEnv } from './core-validator' // your zero-dep lib

export function validateEnvPlugin(schema: any): Plugin {
  let cfg: import('vite').ResolvedConfig

  return {
    name: 'validate-env',
    apply: 'build', // or 'serve' if you want dev checks too
    configResolved(c) {
      cfg = c
    },
    buildStart() {
      const env = loadEnv(cfg.mode, cfg.root, '') // loads .env, .env.[mode], etc.
      const res = validateEnv(schema, env)

      if (res.ok) return

      // Pretty print via Vite's logger (uses picocolors internally)
      const log = cfg.logger
      log.error('\nMissing or invalid environment variables:\n')
      for (const issue of res.issues) {
        log.error(`  • ${issue.key} -> ${issue.code}${issue.detail ? ` (${issue.detail})` : ''}`)
      }
      log.error('\nFix your .env or system env and try again.\n')

      // Fail the build with a Rollup-style error for nice stack-less output
      this.error('Environment validation failed')
    },
  }
}

```

yamcodes:
Closes #206

### DIFF
--- a/packages/arkenv/src/create-env.ts
+++ b/packages/arkenv/src/create-env.ts
@@ -2,6 +2,7 @@ import type { InferType } from "@repo/types";
 import { type distill, type } from "arktype";
 import { ArkEnvError } from "./errors";
 import { $ } from "./scope";
+import type { LoggerStyle } from "./utils";
 
 type RuntimeEnvironment = Record<string, string | undefined>;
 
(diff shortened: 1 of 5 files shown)


yamcodes:
This is what we gain from this:

image

For reference, this is what we have now:

image


I'm not sure we need it. As it appears, it's a whole lot of code for what is essentially a "nice-to-have". I'm not maintaining this. Closing for now.

PR #894, opened 2026-05-08
Add regression tests for inline schema autocompletion

### LINKED ISSUE #796: Add regression tests for inline schema autocompletion
## Context

Inline schema autocompletion (without the `type()` helper) has regressed twice now:

- #522 — first reported in v0.7.7, fixed in v0.7.8 via #531
- #790 — regressed again in v0.9.x

Both times the root cause was changes to the `createEnv` overload signatures in `create-env.ts` that broke TypeScript's ability to provide string literal suggestions for ArkType DSL values.

## Task

Add automated regression tests that catch autocompletion breakage. Options to explore:

- **`@arktype/attest`** or similar type-level testing to assert that `EnvSchema<T>` resolves correctly for inline objects
- **dtslint / tsd / expect-type** style assertions that verify the parameter type narrows to ArkType's string DSL union
- A simple type-level test that ensures `arkenv({ KEY: "..." })` infers the same types as `arkenv(type({ KEY: "..." }))` — if the types diverge, autocompletion likely broke
- Vitest type testing

The goal is to prevent this from silently regressing a third time.

yamcodes:
Closes #796


pnpm-workspace.yaml
```diff
@@ -26,7 +27,9 @@ catalog:
   '@types/react-dom': 19.2.3
   '@vitejs/plugin-react': 6.0.1
   '@vitest/ui': 4.1.5
-  arktype: 2.2.0
+  # arktype is pinned to 2.1.22 for compatibility with @ark/attest@0.56.0
+  # upgrading to 2.2.0 causes "Scope already named Array already exists" errors in type tests
+  arktype: 2.1.22
```
yamcodes: This is weird

packages/arkenv/package.json
```diff
@@ -35,10 +35,11 @@
 	"scripts": {
 		"build": "tsdown",
 		"size": "size-limit --json > .size-limit.json",
-		"test:once": "pnpm test",
+		"test:once": "vitest run",
```
yamcodes, line 38: Not sure this change was neccessary

packages/arkenv/package.json
```diff
@@ -35,10 +35,11 @@
...
 		"typecheck": "tsc --noEmit",
 		"clean": "rimraf dist node_modules",
-		"test": "vitest",
+		"test": "ATTEST_skipTypes=1 vitest",
```
yamcodes: Why `ATTEST_skipTypes=1`?

packages/arkenv/package.json
```diff
@@ -35,10 +35,11 @@
...
+		"test:types": "vitest run src/type-regression.test.ts",
```
yamcodes: This feels a bit too specific, consider using a Vite project or a special extension or something that will consolidate specific type tests without us having to add them here one by one

packages/arkenv/src/type-regression.test-d.ts
```diff
@@ -0,0 +1,88 @@
...
+Overload 2 of 3, '(def: Type<SchemaShape, { string: Submodule<{ trim: Submodule<$ & { " arkInferred": (In: string) => To<string>; }>; normalize: Submodule<$ & { " arkInferred": (In: string) => To<...>; }>; ... 21 more ...; host: string; }>; number: Submodule<...>; }>, config?: ArkEnvConfig | undefined): SchemaShape', gave the following error.Object literal may only specify known properties, and 'KEY' does not exist in type 'Type<SchemaShape, { string: Submodule<{ trim: Submodule<$ & { " arkInferred": (In: string) => To<string>; }>; normalize: Submodule<$ & { " arkInferred": (In: string) => To<...>; }>; ... 21 more ...; host: string; }>; number: Submodule<...>; }>'.
+Overload 3 of 3, '(def: Type<SchemaShape, { string: Submodule<{ trim: Submodule<$ & { " arkInferred": (In: string) => To<string>; }>; normalize: Submodule<$ & { " arkInferred": (In: string) => To<...>; }>; ... 21 more ...; host: string; }>; number: Submodule<...>; }> | validateObjectLiteral<...>, config?: ArkEnvConfig | undefined): SchemaShape | { ...; }', gave the following error.Type '"invalid"' is not assignable to type '"'invalid' is unresolvable "'.`);
+	});
```
yamcodes: Absolutely not

packages/arkenv/package.json
yamcodes: The `ATTEST_skipTypes=1` environment variable is used to **prioritize developer performance** during the inner-loop development cycle.

Here is why it's included in the default `test` script:

### 1. Speed (The Main Reason)
Type-level assertions in `@ark/attest` are incredibly powerful but involve a significant overhead. Every time they run, Attest has to:
*   Spin up or communicate with a **TypeScript Language Service (TSServer)**.
*   Perform static analysis on your source files to verify completions and types.

This can turn a sub-second Vitest run into a 5–10 second run. By setting `ATTEST_skipTypes=1`, you disable this static analysis, making your tests run at near-native speed.

### 2. Development Workflow
During active coding, you usually want instant feedback on whether your logic works (**Runtime validation**). You don't necessarily need to re-verify that `PORT` suggests `"number"` on every single file save.

*   **`pnpm test`**: Runs almost instantly. Validates runtime logic.
*   **`pnpm test:types`**: Takes a few seconds. Validates the complex TypeScript narrowing, completions, and compiler errors.

### 3. How Attest Handles the Flag
When `ATTEST_skipTypes=1` is set, Attest calls like `attest(x).type.errors(...)` are not completely ignored; instead:
*   They **skip the static type check**.
*   They **still run the runtime check**. For example, if you use `.throwsAndHasTypeError()`, it will still verify that the function actually throws at runtime, but it won't check the compiler's diagnostic messages.

### 4. CI Safety
In `package.json`, we updated the `test:once` script (which CI typically calls) to **not** include this flag. This ensures that while developers get a fast local experience, the CI pipeline still performs the full, deep type validation before any code is merged.

PR #932, opened 2026-05-11
docs: replace image tag with `next/image` in import guide for optimization

### LINKED ISSUE #930: Broken image in IDE import guide
The image at `apps/www/content/docs/arkenv/integrations/ide/import-guide.mdx` with path `/assets/dx.png` is not displaying correctly in the documentation site. 

The file exists at `apps/www/public/assets/dx.png`, so the path should be correct for a Next.js public asset, but it needs investigation.

yamcodes:


yamcodes:
@gemini-cli /review

PR #973, opened 2026-05-13
Improve `env.d.ts` handling in CLI

### LINKED ISSUE #972: feat(cli): Improve vite-env.d.ts handling with smart prompts and append mode
### Is your feature request related to a problem? Please describe.

Currently, the `@arkenv/cli` has a frustrating UX when handling the `vite-env.d.ts` file. It exhibits a classic CLI anti-pattern: it asks a hypothetical question ("Establish vite-env.d.ts?"), and if the user says yes but the file already exists, it hits them with a destructive overwrite warning.

This forces developers to manually check their filesystem mid-installation. Furthermore, because we currently only support a hard overwrite, doing so deletes default Vite configurations (like `/// <reference types="vite/client" />`), which breaks standard Vite types.

### Describe the solution you'd like

We need to combine the UX fix and the safe-injection feature into a single PR. The CLI should check the filesystem before prompting the user, and offer a safe "append" option if the file already exists.

### Expected CLI Flow

**Scenario A: `vite-env.d.ts` DOES NOT exist**
```
◇  Establish vite-env.d.ts for typesafe environment variables?
│  Yes (Recommended)
```
*(Action: Create file with ArkEnv types)*

**Scenario B: `vite-env.d.ts` ALREADY exists**
```
◇  Found existing vite-env.d.ts. How should we handle ArkEnv types?
│  Append types safely (Recommended)
│  Overwrite entirely
│  Skip
```
*(Action: Append the required ArkEnv references to the end/top of the file without touching existing Vite types)*

### Technical Tasks / Checklist

- [ ] Add `fs.existsSync` check for `vite-env.d.ts` prior to the prompt step.
- [ ] Refactor the prompt logic to branch conditionally based on the file's existence.
- [ ] Implement the append logic to safely inject the ArkEnv `/// <reference ... />` or interface into the existing file.
- [ ] Ensure the prompt text matches the new branching logic so users are never asked twice about the same file.
- [ ] Update any relevant CLI tests to cover both the "file exists" and "file missing" scenarios.

### Additional context

We want to maintain the "don't make me think" philosophy. Defaulting to safe appends ensures we don't accidentally nuke a user's existing Vite or SVGR type setups.

yamcodes:
Closes #972

## Summary by CodeRabbit

* **New Features**
  * Safe-injection to append ArkEnv type definitions (no overwrite) and a new option to skip/append/overwrite during scaffolding.
  * Scaffolding now reports an "appended" status when types are appended.

* **Documentation**
  * Added design, proposal, and spec docs describing safe-injection, prompt flows, and CLI behavior.

* **Tests**
  * Added tests covering safe append, skip/overwrite/append scenarios, and path-handling edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/yamcodes/arkenv/pull/973)


yamcodes:
Plan looks good

PR #976, opened 2026-05-14
966 pnpm skill

### LINKED ISSUE #966: pnpm skill
The CLI keeps trying to use npm, add a pnpm skill

yamcodes:
Fixes #966


yamcodes:
Not a fan of those global no null assertions and no any rule overrides.

yamcodes:
Actually I'm sold. We don't need to check for "any" in test files and we don't need strict null checks in playgrounds/examples when we're trying to prototype

PR #1042, opened 2026-05-21
feat: force new project flow if --example is passed

### LINKED ISSUE #1040: Force "new project flow" if --example is passed
Even if the current directory is not empty, force the "new project flow" if --example is passed. in that case, still prompt for the project name and continue from there. Only if the provided project name is "." and the current dir is not empty, THEN abort with "non empty" failure.

yamcodes:
Fixes #1040

## Summary

When `--example` is passed, the CLI now always enters the new-project wizard, regardless of whether the directory is non-empty or already has a `package.json`. Previously the existing-project / non-empty-directory guards ran first, silently ignoring the flag.

## Changes

- **`InitUseCase.collect()`** (): Added an early guard — if `input.example !== undefined`, skip straight to `collectNewProject()` before the existing-project / empty-directory checks.
- **`InitUseCase.collectNewProject()`**: Accepts an `isEmpty` flag. After the wizard returns, aborts with a clear error if `options.name === "."` and the directory was not empty; normalizes `"."` to the directory basename only when the directory is empty.
- **`runNewProjectWizard()`** (`packages/cli/src/cli/ui/prompts/wizard.ts`): Removed the premature `"."` → `path.basename(cwd)` conversion so the raw value is preserved through to `collectNewProject()` where the guard lives.
- **Tests**: 3 new `InitUseCase` tests and 1 new `runPromptWizard` test covering all described scenarios.

### DIFF
--- a/.changeset/example-forces-new-project-flow.md
+++ b/.changeset/example-forces-new-project-flow.md
@@ -1,0 +1,19 @@
+---
+"@arkenv/cli": minor
+---
+
+#### `--example` now forces the new-project wizard regardless of the current directory
+
+Previously, passing `--example` in a non-empty directory (or one that already has a
+`package.json`) would silently fall through to the existing-project flow, ignoring the
+flag entirely. The flag now always triggers the new-project wizard:
+
+```sh
+# Works even in a non-empty directory or one with package.json
+arkenv init --example basic
+```
+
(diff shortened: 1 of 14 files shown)


yamcodes:
@pullfrog go ahead and fix the merge conflicts. The main branch changed the `--name` flag to an optional positional argument. We need that reflected here, in the code and in the changeset.

PR #1060, opened 2026-05-21
refactor(cli): modularize prompt wizard and steps into pure type safe modules

### LINKED ISSUE #1057: refactor(cli): modularize prompt wizard and steps into pure, type-safe modules
### Problem
The interactive CLI onboarding prompt flow in `@arkenv/cli` is currently spread across 8 separate modules. While they are separate files, the design is highly coupled and shallow:
1. They communicate using an untyped `{ results: any }` context object.
2. The orchestrator in `wizard.ts` coordinates them via a dynamic array of functions using complex inline ternary checks (e.g. for `bunBuild`), which makes static type-checking impossible.
3. Steps like `env-example.ts` and `env-types.ts` bypass the `WorkspacePort` seam by importing `node:fs` directly to check for file existence, forcing tests to mock Node's global `fs` module.

### Proposed Solution
Refactor the prompt wizard and step modules using a hybrid approach (**Option A**):
1. **Sequential Orchestration**: Replace the dynamic loop in `wizard.ts` with a direct, type-safe sequential pipeline of step functions.
2. **Pure Step Functions**: Refactor each step module in `steps/` to be a pure function that takes explicit typed inputs and returns typed results (or `null` on cancellation).
3. **Clean Seams**: Remove direct `node:fs` calls. Instead, scan the workspace using `WorkspacePort` in the orchestrator/caller and pass the boolean file existence flags down as arguments.
4. **Cohesion**: Keep all individual modules well under 200 lines to avoid creating an "everything" file.

### Benefits
- **Locality**: The sequencing of wizard prompts is readable in a single screen inside `wizard.ts`.
- **Depth**: The `runPromptWizard` interface completely hides the complexity of prompt execution, and the individual steps have precise, simple interfaces.
- **Type-Safety**: Static analysis works perfectly across steps.
- **Testability**: Tests are simplified because they no longer need to mock the global `fs` module.

yamcodes:


yamcodes (commented):
LGTM but I'd run the /jsdoc skill across every file changed.

PR #1061, opened 2026-05-21
fix: Cannot scaffold into a non-empty directory despite using --force

### LINKED ISSUE #1053: Cannot scaffold into a non-empty directory despite using `--force`
See:

```sh
code/arkenv-tests/bun-test 
❯ pnx @arkenv/cli@0.2.1 init                                     
Packages: +17
+++++++++++++++++
Progress: resolved 17, reused 16, downloaded 1, added 1, done
✘ Directory is not empty and no package.json was found.
ℹ To scaffold a new project, run arkenv init in an empty directory or use --force to proceed anyway.

code/arkenv-tests/bun-test took 3s 
❯ pnx @arkenv/cli@0.2.1 init --force
│
◇  Project name:
│  .
│
◇  Select an example:
│  Bun (Vanilla)
✘ Cannot scaffold into . because the current directory is not empty.
ℹ Run arkenv init in an empty directory or choose a sub-directory name instead.
```

I would have expected to be able to continue with the scaffold or at least have a clear indication right from the get-go that it would be impossible to do so and an explanation why.

yamcodes:
Fixes #1053

Allows scaffolding a new project into a non-empty directory (such as '.') when the '--force' flag is specified. Implements a preflight copy check to ensure that no existing user files or directories are silently overwritten.

### DIFF
--- a/.changeset/allow-force-scaffold-dot.md
+++ b/.changeset/allow-force-scaffold-dot.md
@@ -1,0 +1,7 @@
+---
+"@arkenv/cli": patch
+---
+
+#### Allow scaffolding into non-empty directory when `--force` is used
+
+Ensure `--force` permits new-project scaffolding into a non-empty directory (e.g. `.`) while preventing silent overwrites of user files via a preflight collision check.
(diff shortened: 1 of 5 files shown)


yamcodes:
@pullfrog all good ?

PR #1087, opened 2026-05-28
fix: Update Vite peerDependency floor to match tested versions

### LINKED ISSUE #864: Test older Vite versions
Right now we have a ["vite-legacy" playground](https://github.com/yamcodes/arkenv/tree/main/apps/playgrounds/vite-legacy) to softly back up the peer dependency range declared in `@arkenv/vite-plugin`:

```json
"vite": "^2.9.18 || ^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0"
```

## Current state

- `vite-legacy` is pinned to **Vite 7** — it builds as part of `pnpm run build` via Turbo, so CI does verify at least one non-latest major.
- The Vite 2–6 range in `peerDependencies` is **untested** and unverified.

## Long-term plan

Add a CI matrix job (in `test.yml`) that installs and builds `vite-legacy` against each claimed major:

```yaml
strategy:
  matrix:
    vite-version: ["4", "5", "6", "7"]
```

Each job would override the `vite` dep (e.g. via `pnpm --override vite@${{ matrix.vite-version }}`) and run `pnpm run build --filter=vite-legacy-playground`.

Until that matrix exists, the `peerDependencies` floor should stay at `^7.0.0` (the lowest actually tested version) to avoid false promises to users on older Vite versions. Once the matrix is in place, the floor can be lowered only as far as the matrix validates.

yamcodes:
Fixes #864

This PR adds a follow-up to the merged #1086:
1. Restricts the `peerDependencies` of `vite` in `@arkenv/vite-plugin` to start from `^4.0.0` (reflecting only the versions validated in our CI compatibility matrix).
2. Refactors `scripts/override-vite.js` to add support for a `--restore` flag, preventing dirty workspace states locally and avoiding duplicate entries when run multiple times.
3. Adds a changeset file documenting the peer dependency change.


yamcodes:
@pullfrog please update the docs to reflect this new change (just modify the ranges where it says them)

PR #1088, opened 2026-05-28
feat: add withArkEnv config generator to auto-generate runtime enviro…

### LINKED ISSUE #1072: feat(nextjs): codegen for runtimeEnv destructuring
### Describe the Feature Request

Currently, developers using `@arkenv/nextjs` must manually destructure all client and shared environment variables in `runtimeEnv` to satisfy Next.js's AST compiler inlining requirements (which replaces `process.env.NEXT_PUBLIC_...` statically in client bundles):

```ts
export const env = arkenv({
  client: {
    NEXT_PUBLIC_API_URL: "string",
  },
  runtimeEnv: {
  }
});
```

This introduces unnecessary boilerplate and requires keeping the schema and `runtimeEnv` block in sync.

### Proposed Solution

Implement automatic code generation for the `runtimeEnv` block so developers do not have to write it manually. 

We can expose a config wrapper/helper from `@arkenv/nextjs/config` that runs programmatically when `next.config.js` is loaded:

```javascript
// next.config.js
const { withArkEnv } = require("@arkenv/nextjs/config");

module.exports = withArkEnv({
  // NextConfig options
});
```

When Next.js runs (`next dev` or `next build`), `withArkEnv` will:
1. Locate the source schema configuration (e.g. defined in `env.config.ts`).
2. Read the client/shared keys.
3. Automatically generate and write `env.ts` on disk containing the compiled `runtimeEnv` block.
4. Pass the config along.

### Alternative Implementation: CLI-based Codegen

As an alternative, this could be implemented as a CLI command (e.g. `@arkenv/cli generate --watch`) that parses `env.config.ts` and writes out `env.ts`. However, the `next.config.js` wrapper is preferred as it is completely automatic and requires no extra watcher scripts running in parallel with `next dev`.

yamcodes:
…nment schemas


yamcodes:
This came out wrong in almost every way, closing

PR #1090, opened 2026-05-29
fix: ArkEnv skill should be detected by the CLI

### LINKED ISSUE #1089: ArkEnv skill should be detected by the CLI
# Refined Specification

## Problem Statement
When running the `@arkenv/cli` `init` command in an existing project, the CLI prompts the user to install the ArkEnv agent skill:
`Would you like to install the ArkEnv agent skill?`
If the user already has the ArkEnv agent skill installed in their project (e.g. in `skills/arkenv`, `.agent/skills/arkenv`, `.agents/skills/arkenv`, or configured in `skills-lock.json`), this prompt is redundant and confusing, and proceeding with it attempts to run the installation process again.

## Desired Behavior
1. The CLI should automatically detect if the `arkenv` skill is already installed.
2. The detection logic should look for:
   - A `skills-lock.json` file in the target directory (the project root), and check if the `skills` object contains the key `"arkenv"`.
   - The existence of a skill definition file (`SKILL.md`) at any of the following paths:
     - `skills/arkenv/SKILL.md`
     - `.agent/skills/arkenv/SKILL.md`
     - `.agents/skills/arkenv/SKILL.md`
3. If the skill is detected:
   - In interactive mode, the prompt asking `Would you like to install the ArkEnv agent skill?` should be bypassed, and `options.installSkill` should default to `false`.
   - In non-interactive mode (`--yes`), `options.installSkill` should default to `false` instead of `true`.
   - The CLI should print an informational message or log confirming detection: `ArkEnv agent skill detected.` (unless running in quiet/agent mode).

## Acceptance Criteria
- [ ] Implement a skill detection utility in the CLI package.
- [ ] Add unit/integration tests verifying the detection logic with mock project structures (e.g. mock `skills-lock.json` or directory paths).
- [ ] Ensure that `arkenv init` does not prompt to install the skill when it is detected.
- [ ] Ensure that `arkenv init --yes` does not try to install the skill when it is detected.
- [ ] Ensure that a detection message is logged to the output when detected in non-quiet mode.

yamcodes:
Fixes #1089

Automatically detects if the arkenv AI skill is already installed (checks skills-lock.json and SKILL.md paths) and skips the installer prompts/installation steps during arkenv init.

### DIFF
--- a/.changeset/detect-existing-skill.md
+++ b/.changeset/detect-existing-skill.md
@@ -1,5 +1,5 @@
 ---
-"@arkenv/cli": minor
+"@arkenv/cli": patch
 ---
 
 #### Automatically detect installed arkenv AI skill
--- a/.changeset/detect-existing-skill.md
+++ b/.changeset/detect-existing-skill.md
@@ -1,0 +1,7 @@
+---
+"@arkenv/cli": minor
+---
+
+#### Automatically detect installed arkenv AI skill
+
+During initialization, check if the `arkenv` agent skill is already present in the workspace. If detected, the installation prompt and setup are bypassed, defaulting to `false`, and an informational message confirming the detection is logged.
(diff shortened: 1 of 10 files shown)


.changeset/detect-existing-skill.md
```diff
@@ -0,0 +1,7 @@
+---
+"@arkenv/cli": minor
+---
```
yamcodes, line 3: This should be `patch` - we're in v0

PR #1092, opened 2026-05-30
feat(nextjs): codegen for runtimeEnv destructuring

### LINKED ISSUE #1072: feat(nextjs): codegen for runtimeEnv destructuring

yamcodes:
Fixes #1072

Adds a new codegen helper `withArkEnv()` inside `@arkenv/nextjs/config` that runs during Next.js config loading. It automatically extracts keys from your schema (`env.ts`) and generates a tailored `createEnv` factory inside `env.gen.ts` that pre-fills the `runtimeEnv` block.

This removes the boilerplate and the need to manually destructure environment variables to satisfy Next.js's static client inlining rules.


yamcodes:
We need a machine code review here.

yamcodes:
This is good but I would like to explore some alternatives before committing to a codegen solution

yamcodes:
Code review from `Claude Opus 4.6 (Thinking)`:

[code_review_pr1092.md](https://github.com/user-attachments/files/28417709/code_review_pr1092.md)

Here's my review of PR #1092. The full details are in the artifact above — here's the executive summary:

**CI Status**: ✅ All 21 checks passing

## Key Findings

### 🔴 3 Blocking Issues

1. **`@repo/scope` hardcoded in generated output** — The `generateFactoryCode` template imports `@repo/scope`, a private monorepo package. Any user installing `@arkenv/nextjs` from npm will get a broken `env.gen.ts` that can't compile.

2. **`extractBlock` ignores braces inside strings/comments** — It counts `{`/`}` naively. A schema value like `"string = '{pattern}'"` would break the brace counter and corrupt key extraction.

3. **`parseBlockKeys` can capture spurious keys from nested values** — An ArkType expression like `type("string", { description: "foo" })` would cause `description` to be extracted as an env var key.

### 🟡 8 Non-Blocking Suggestions

- Deprecating the entire existing API in a **patch** release is aggressive — consider a minor bump or deferring deprecation
- `fs.watch` is unreliable cross-platform — needs debouncing at minimum  
- Stream-of-consciousness comments in tests ("Oh, wow! This is so beautiful!") should be removed
- DRY violation between `getUsageInstructions` and `getNextStepsNote`
- Several other documentation and config hygiene items

### ✅ Praise

The overall architecture is elegant — the generated factory pattern is completely decoupled from the schema file at runtime. Write-if-changed optimization, comprehensive test coverage, and consistent CLI scaffold updates are all well done.

yamcodes:
Antigravity:

> I have reviewed the code. Everything looks great, but I found one issue where the parser wouldn't recognize ArkType `type({ ... })` wrappers, which would break support for the newly unified `Infer<T>` API changes. I've gone ahead and pushed a fix that handles optional wrappers and added tests. Other than that, the codegen logic is brilliant and works perfectly!

yamcodes:
hmm, we're getting `Do not use "@ts-nocheck" because it alters compilation errors.eslint[@typescript-eslint/ban-ts-comment](https://typescript-eslint.io/rules/ban-ts-comment)`

what's the best practice here?

yamcodes:
i was not asked for the "generated" file location. i would have preferred it at `generated/` folder. maybe we should add that as a question, and maybe it should even be the new default. not sure what's the convention here

yamcodes:
after finishing `pnpm arkenv init`, we are in an invalid state, because in `env.ts` we have:

```ts
import { createEnv } from "./env.gen";

/**
 * Environment variable schema.
 * In Next.js, use the generated `createEnv` from `env.gen.ts` to validate variables.
 * Enforces client/server separation and prevents secret leaks.
 */
export const env = createEnv({
	server: {
		DATABASE_URL: "string = 'postgres://localhost:5432/mydb'",
	},
	client: {
		NEXT_PUBLIC_API_URL: "string = 'https://api.example.com'",
	},
	shared: {
		NODE_ENV: "'development' | 'production' | 'test' = 'development'",
	},
});


```
But there's no `env.gen` file yet.

I would expect us to "force" a first-generation so that the first usage (even before `pnpm run dev`) will work cleanly.

---

```ts
◇  Next steps ────────────────────────────────────────────────────────────╮
│                                                                         │
│  1. Check ./env.ts and refine your environment schema.                  │
│  2. Wrap your Next.js config with withArkEnv inside next.config.ts:     │
│     import { withArkEnv } from "@arkenv/nextjs/config";                 │
│     export default withArkEnv(nextConfig);                              │
│  3. Import and use: import { env } from "./env"                         │
│  4. (Recommended) Install the AI skill: npx skills add yamcodes/arkenv  │
│     Then run /arkenv inside your AI assistant to finish.                │
│                                                                         │
├─────────────────────────────────────────────────────────────────────────╯
```

The next steps say to wrap our config with `withArkEnv` but... why isn't the CLI doing it for us? (This might be out of scope for this iteration - it's ok to open a new issue for this instead of implementing here.)

yamcodes:
`outputPath` and `schemaPath` should have their own `jsdoc` with clear examples

yamcodes:
LGTM

yamcodes:
Some people don't like codegen, we should have a way to opt out like in the CLI.

yamcodes:
Needs final QA but we're practically there

yamcodes:
Why -C ?

.changeset/disable-codegen-option.md
```diff
@@ -0,0 +1,9 @@
+---
+"@arkenv/cli": patch
+---
+
+#### Add `--no-codegen` CLI option and dedicated prompt for Next.js scaffolding
+
+Introduce a `--no-codegen` (or `-C`) option and an interactive prompt to allow developers to opt out of the Next.js automatic environment variable code generation workflow. When opted out, the CLI scaffolds the project to use standard runtimeEnv destructuring and skips post-scaffold code generation bootstrapping.
+
+Additionally, fix a TypeScript compilation check error in the CLI planner by adding the missing `parsed` property type to `CollectedState["tsConfig"]`.
```
yamcodes: This is unnecessary

.changeset/disable-codegen-option.md
yamcodes, line 1: this whole file can be combined with the other CLI's changeset

packages/cli/src/features/scaffold/utils.test.ts
yamcodes, line 1: why was this file removed?

Issue #1112, opened 2026-05-31
clean up nextjs docs


yamcodes:
the nextjs docs at https://arkenv.js.org/docs/nextjs should be cleaned up a bit - to be written in a more human form, and split in a sane way. currently the page is just too busy and things are not rendering correctly.

also the happy path with arkenv cli is not visible enough

PR #1135, opened 2026-06-01
fix: restrict shared variables in Next.js to NODE_ENV

### LINKED ISSUE #1131: Restrict `shared` variables in Next.js to `NODE_ENV` to prevent hydration mismatches
### Problem Statement
In Next.js, accessing a custom environment variable in a Client Component that does not have the `NEXT_PUBLIC_` prefix causes the Next.js compiler/bundler to strip it from the client-side bundle, resulting in `undefined` at runtime.
When using `arkenv`, if a user places a custom environment variable in the `shared` block of the schema (which allows client-side access and bypasses the `NEXT_PUBLIC_` prefix requirement) and defines a default fallback (e.g., `string = 'hello'`), the server will resolve the real environment variable value, while the client will resolve it to the default value (`hello`).

This "Environment Asymmetry" creates a silent developer experience (DX) trap resulting in React hydration mismatches and corrupt application state.

#### Why this only affects `shared` variables:
1. **`server` variables**: If accessed on the client, the Proxy intercepts and immediately throws an error: `Accessing server-side environment variable 'MY_ENV' on the client is not allowed.` This prevents mismatched values.
2. **`client` variables**: Strictly validated to require the `NEXT_PUBLIC_` prefix. Next.js guarantees they are statically inlined on both server and client, ensuring consistency.
3. **`shared` variables**: Act as a double-bypass, escaping both the server-side proxy check and the client-side prefix validation.

### Desired Behavior
1. **CLI Scaffolding/Templates**: Update all CLI nextjs templates to only include `NODE_ENV` in the `shared` block and treat `PORT` (or other custom variables) as `server`-only.
2. **Tests**: Update all scaffold, nextjs config, and regression tests to expect `PORT` in the server block and only `NODE_ENV` in the shared block.
3. **Documentation**: Add a "Gotchas / Danger of Shared Variables" section in the Next.js documentation (e.g., package READMEs and docs site) explaining this static bundler mechanics edge case.

### Acceptance Criteria
- [ ] Next.js CLI templates only contain `NODE_ENV` in the `shared` block; custom/port variables are generated as `server`-only.
- [ ] JSDoc examples and docs do not recommend placing custom variables in `shared` without `NEXT_PUBLIC_` prefix.
- [ ] Next.js integration docs are updated with a dedicated "The Danger of Shared Variables" warning section detailing the "Undefined Fallback" hydration mismatch bug.
- [ ] All unit, integration, and type tests are updated and pass successfully.

### Strong example

```ts
// ❌ DANGER: Asymmetric Shared Variable
export const env = createEnv({
  shared: { MY_THEME: "string = 'dark'" } 
});

// In a Client Component:
export function ThemeToggle() {
  // Server renders: "light" (from real process.env)
  // Client hydrates: "dark" (Next.js strips process.env, arkenv falls back to default)
  // Result: React Hydration Error!
  return <div>{env.MY_THEME}</div>; 
}
```

yamcodes:
Fixes #1131

Restrict Next.js environment configuration templates and scaffolding generated by the CLI so that custom variables (like `PORT`) are generated as server-only variables instead of shared variables. This prevents client-side hydration mismatches since Next.js strips non-prefixed variables on the client.

### DIFF
--- a/apps/www/content/docs/nextjs/index.mdx
+++ b/apps/www/content/docs/nextjs/index.mdx
@@ -457,3 +457,20 @@ You can freely mix Zod or Valibot schemas with the 3-file layout:
     ```
   </Tab>
 </Tabs>
+
+---
+
+## The Danger of Shared Variables
+
+> [!WARNING]
+> Restrict the `shared` block only to `NODE_ENV`. Avoid placing custom variables (like `PORT` or other custom configuration) in the `shared` block.
+
+### The Undefined Fallback Bug
+
+Next.js statically strips `process.env` references from client-side bundles unless they are prefixed with `NEXT_PUBLIC_`.
+If you define a custom variable in `shared` with a default value (e.g., `PORT` defaulting to `3000` or `THEME` defaulting to `'dark'`), the environment behaves asymmetrically:
+
+- On the **server**, ArkEnv reads the actual value from the environment (e.g. `PORT = 8080`).
+- On the **client**, Next.js strips `process.env.PORT` to `undefined`, causing ArkEnv to fall back to the default value (`3000`).
+
+This asymmetry causes React hydration mismatches and corrupts client-side state. Always place non-prefixed variables under `server` so they are strictly server-only.
--- a/packages/cli/src/features/scaffold/env-template.ts
+++ b/packages/cli/src/features/scaffold/env-template.ts
@@ -80,28 +80,42 @@ export function getStrictEnvTemplates(
 					clientFields.push(`${key}: v.optional(v.string()),`);
 				}
 				runtimeEnvFields.push(`${key}: process.env.${key},`);
-			} else if (key === "NODE_ENV" || key === "PORT") {
+			} else if (key === "NODE_ENV") {
 				if (validator === "arktype") {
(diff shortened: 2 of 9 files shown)


apps/www/content/docs/nextjs/index.mdx
yamcodes, line 1: Not really happy with the way this is written, also if it's gonna be a copy-paste of the nextjs' README we could just reference it from there like we once did in the "start with an example" page (see git historry there)

packages/cli/src/features/scaffold/env-template.ts
yamcodes, line 1: I never noticed this but this code is written in a very ugly way

yamcodes (commented):
no blocking changes

packages/cli/src/features/scaffold/env-template.ts
yamcodes: @pullfrog can you open a followup issue for this, so many nested if's

apps/www/content/docs/nextjs/index.mdx
yamcodes: needs to be fixed in #1112

PR #1139, opened 2026-06-01
fix: remove deprecation from createEnv export in @arkenv/nextjs

### LINKED ISSUE #1128: why is `createEnv` export deprecated in "@arkenv/nextjs"
it shouldn't be deprecated because codegen is optional

yamcodes:
Fixes #1128

Remove the incorrect @deprecated JSDoc tag from createEnv and arkenv in the main and react-server entry points of @arkenv/nextjs.


yamcodes (commented):
LGTM

PR #1156, opened 2026-06-03
infra: migrate default branch to dev

### LINKED ISSUE #1043: www can document unreleased features
Becuase of the way our workflows work (Changesets being one thing, and deploys to vercel when hitting main being another), the www website (in production) can document unreleased features. Only when the "Version Packages" PR is merged to main is the info on www accurate.

This creates gap of mis-documentation that ultimately cause pressure to release faster, which might lead to bugs and is more wasteful.

yamcodes:
Closes #1043. Migrates default branch configuration, workflow targets, changesets, and docs to target `dev`.

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -3,7 +3,7 @@ name: autofix.ci # needed to securely identify the workflow
 on:
   pull_request:
   push:
-    branches: ["main"]
+    branches: ["dev"]
 permissions:
   contents: read
 
@@ -40,7 +40,7 @@ jobs:
         run: |
           SKIP_MANYPKG=false
 
-          if [ "${{ github.event_name }}" = "pull_request" ] && [ "$HEAD_REF" = "changeset-release/main" ]; then
+          if [ "${{ github.event_name }}" = "pull_request" ] && [ "$HEAD_REF" = "changeset-release/dev" ]; then
             SKIP_MANYPKG=true
           elif [ "${{ github.event_name }}" = "push" ]; then
             if echo "$COMMIT_MESSAGE" | grep -qi "Version Packages"; then
--- a/.github/workflows/sync-contributors.yml
+++ b/.github/workflows/sync-contributors.yml
@@ -5,7 +5,7 @@ on:
     paths:
       - ".all-contributorsrc"
   push:
-    branches: ["main"]
+    branches: ["dev"]
     paths:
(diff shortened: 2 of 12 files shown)


.github/workflows/autofix.yml
```diff
@@ -3,7 +3,7 @@ name: autofix.ci # needed to securely identify the workflow
 on:
   pull_request:
   push:
-    branches: ["main"]
+    branches: ["dev"]
```
yamcodes, line 6: only dev or also main. ?

.github/workflows/sync-contributors.yml
```diff
@@ -5,7 +5,7 @@ on:
     paths:
       - ".all-contributorsrc"
   push:
-    branches: ["main"]
+    branches: ["dev"]
```
yamcodes, line 8: dev, main or both ?

.github/workflows/sync-examples.yml
```diff
@@ -8,7 +8,7 @@ on:
       - "scripts/sync-examples.js"
       - "scripts/sync-lib/**"
   push:
-    branches: ["main"]
+    branches: ["dev"]
```
yamcodes, line 11: dev , main or both ?

.github/workflows/test.yml
yamcodes, line 1: dev main or both?

.github/workflows/autofix.yml
yamcodes: Only `dev`. Since `main` is strictly updated via a fast-forward merge from `dev`, running autofix on `main` would be redundant compute.

.github/workflows/sync-contributors.yml
yamcodes: Only `dev`. If this ran on `main`, it could commit updates directly to `main`, causing it to drift from `dev` and breaking subsequent fast-forward releases.

.github/workflows/sync-examples.yml
yamcodes: Only `dev`. Just like the contributors sync, we want to prevent direct commits to `main` so it never drifts from `dev`.

.github/workflows/test.yml
yamcodes: Only `dev`. Since `main` is a byte-for-byte copy of the already-verified `dev` branch, re-running tests on `main` is redundant compute and risks flaky tests blocking our deployments.

PR #1193, opened 2026-06-12
fix: Simplify coercion: unify ArkType and Standard Schema paths with pre-coercion model

### LINKED ISSUE #1178: Simplify coercion: unify ArkType and Standard Schema paths with pre-coercion model
## Problem

ArkEnv has two different execution models for environment variable coercion:

1. **ArkType path** (`arkenv`): Wraps the schema in an ArkType `.pipe()` pipeline — `at("unknown").pipe(coerce).pipe(schema)`. This composes coercion as a morph *inside* ArkType's type system, which requires an unsafe `as BaseType<t, $>` cast due to scope mismatch.

2. **Standard Schema path** (`arkenv/standard`): Pre-coerces a shallow copy of the env object, then validates the coerced data against the original validator.

Both paths already share the same core coercion logic (`findCoercionPaths` + `applyCoercion` in `coercion/shared.ts` and morph functions in `coercion/morphs.ts`). The asymmetry is only in the thin ArkType-specific wrapper (`arktype/coercion/coerce.ts`) that uses `.pipe()`.

## Desired behavior

Replace the ArkType `.pipe()` wrapper with the same pre-coercion model used by `arkenv/standard`:

```
env (raw strings)
  → shallow copy
  → applyCoercion(copy, targets)
  → schema(coercedCopy)
  → result or errors
```

Both entry points should follow this identical execution flow. The ArkType path still uses `schema.in.toJsonSchema({ fallback })` for JSON Schema extraction (preserving the unjsonifiable fallback), while Standard Schema uses `validator["~standard"].jsonSchema.input()`.

## Why this direction

- **Pre-coercion is the only model that works for both paths.** Standard Schema validators are opaque (`validate(input) → result`); you cannot inject a coercion step inside them. The `.pipe()` pattern is ArkType-specific and can't be generalized.
- **The `.pipe()` approach is less typesafe in practice** — it requires an unsafe `as BaseType<t, $>` cast to work around the scope mismatch between `at("unknown")` (scope `{}`) and the user's schema (scope `$`).
- **The `.pipe()` approach mutates `process.env` in-place** because `applyCoercion` mutates its input and ArkType passes the reference through the pipeline. Pre-coercion with `{ ...env }` fixes this.
- **No native ArkType coercion is planned** (confirmed with David), so keeping the `.pipe()` wrapper as a migration bridge is unnecessary.
- **The ArkType-idiomatic tradeoff** (composing transformations as type pipelines) is real but not worth the architectural complexity.
- **The unjsonifiable fallback does not need to be ported to Standard Schema.** Zod and Valibot produce good JSON Schema for all common types, including refined ones. The fallback is mainly needed for exotic ArkType features (custom morphs, predicates) that Standard Schema validators don't have.

### Code changes
- Replace the `.pipe()` wrapper in the ArkType `parse()` function with direct pre-coercion using `findCoercionPaths` + `applyCoercion` from `coercion/shared.ts`
- Delete `arktype/coercion/coerce.ts` (and its re-export in `arktype/coercion/index.ts`) once the wrapper is no longer needed
- Update unit tests in `arktype/coercion/coerce.test.ts` — the coercion unit tests should move to test `findCoercionPaths` + `applyCoercion` from the shared module directly (or may already exist there)
- Ensure all existing integration tests (`coercion.integration.test.ts`) still pass

### Documentation changes
- **Coercion docs** (`coercion.mdx`): Document the unified pre-coercion model. Explain how ArkEnv introspects JSON Schema, pre-coerces the env object, and validates. Mention that this model is shared across both `arkenv` and `arkenv/standard`.
- **FAQ** (`faq.mdx`): Add or update entries explaining:
  - How coercion works under the hood (JSON Schema introspection → targeted pre-coercion → validation)
  - That coercion is now available in `arkenv/standard` too (not just ArkType)
  - The difference between ArkType coercion and Standard Schema coercion (ArkType has the unjsonifiable fallback for exotic types; Standard Schema relies on the validator's JSON Schema output)
- **Standard Schema docs** (`standard.mdx` / `standard-schema.mdx`): Update the coercion comparison table and any notes that say coercion is ArkType-only
- **ADR-0002** (`0002-coercion-schema-transformer.md`): Update to reflect the architectural change from `.pipe()` to pre-coercion, linking to this issue for context

## Non-goals
- Changing the coercion behavior itself (which types are coerced, how morphs work)
- Porting the unjsonifiable fallback to Standard Schema
- Adding new coercion targets (e.g. dates are already handled on `origin/dev`)
- Changing the public API surface (`coerce`, `arrayFormat` options remain the same)

yamcodes:
Fixes #1178

Unifies the coercion execution model between the default `arkenv` and `arkenv/standard` entry points. Uses a pre-coercion flow that shallow-copies the env object and applies coercion prior to validation against the unmodified schema.

### Changes
- Replace ArkType's `.pipe()` coercion wrapper with pre-coercion flow using `findCoercionPaths` + `applyCoercion` directly in `packages/arkenv/src/arktype/index.ts`
- Delete the redundant wrapper file `packages/arkenv/src/arktype/coercion/coerce.ts` and its re-export
- Update `coerce.test.ts` unit tests to use a local helper mapping the pre-coercion flow
- Update coercion documentation, FAQ, Standard Schema docs, and ADR-0002


packages/arkenv/src/arktype/index.ts
```diff
@@ -96,19 +99,28 @@ export function parse<const T extends SchemaShape>(
...
 	const processedEnv = emptyAsUndefined
 		? stripEmptyStrings(env as Record<string, string | undefined>)
 		: env;
 
+	let coercedEnv = { ...processedEnv } as Record<string, unknown>;
+
+	// Apply coercion transformation to allow strings to be parsed as numbers/booleans
+	if (shouldCoerce) {
+		const json = schemaWithKeys.in.toJsonSchema({
+			fallback: (ctx: any) => ctx.base,
```
yamcodes: Any way to avoid the "any" here?

packages/arkenv/src/arktype/index.ts
```diff
@@ -96,19 +99,28 @@ export function parse<const T extends SchemaShape>(
...
+		});
+		const targets = findCoercionPaths(json as any);
+		if (targets.length > 0) {
+			coercedEnv = applyCoercion(coercedEnv, targets, {
+				arrayFormat,
+			}) as Record<string, unknown>;
```
yamcodes: Any way to avoid this type assertion?

packages/arkenv/src/arktype/index.ts
```diff
@@ -96,19 +99,28 @@ export function parse<const T extends SchemaShape>(
...
 		: env;
 
+	let coercedEnv = { ...processedEnv } as Record<string, unknown>;
+
+	// Apply coercion transformation to allow strings to be parsed as numbers/booleans
+	if (shouldCoerce) {
+		const json = schemaWithKeys.in.toJsonSchema({
+			fallback: (ctx: any) => ctx.base,
+		});
+		const targets = findCoercionPaths(json as any);
```
yamcodes: Any way to avoid this type assertion?

packages/arkenv/src/arktype/index.ts
```diff
@@ -2,11 +2,14 @@ import { $ } from "@repo/scope";
 import type { SchemaShape } from "@repo/types";
 import type { distill } from "arktype";
 import { ArkErrors } from "arktype";
-import { stripEmptyStrings } from "@/coercion/shared";
+import {
+	applyCoercion,
+	findCoercionPaths,
+	stripEmptyStrings,
+} from "@/coercion/shared";
```
yamcodes: I'm wondering if this should stay as "@/coercion/shared" or if we could simplify this import to "@/coercion" - whether with barrel exports or via flattening the tree.

yamcodes (commented):
Just a few /typescript questions but it's good 

I'd also ensure a /jsdoc pass and double check the /changeset although that looked good to me

PR #1218, opened 2026-06-19
fix(nextjs): Next.js integration improvements (DX setup, dynamic client runtime variables, and shared parser refactoring)

### LINKED ISSUE #1145: ArkEnv "flat" mode for Next.js
## The Problem
Currently, the `@arkenv/nextjs` integration requires developers to explicitly separate their environment variables into server, client, and shared blocks.

**Current API (Strict Layout):**
```ts
import arkenv from "@/generated/env.gen";

export const env = arkenv({
  server: {
    DATABASE_URL: "string",
    PORT: "string?",
  },
  client: {
    NEXT_PUBLIC_TEST: "string",
  },
  shared: {
    NODE_ENV: "'development' | 'production' | 'test' = 'development'",
  },
});
```

While this Strict Layout provides explicit architectural boundaries (compile-time locking via Next.js conditional exports), it introduces unnecessary boilerplate and friction for average use cases. Developers typically copy-paste their `.env` files directly, and forcing them to manually categorize variables degrades the Developer Experience (DX). Furthermore, Next.js already enforces a strict `NEXT_PUBLIC_` naming convention for client variables, making explicit client blocks redundant.

## The Proposed Solution (The "Flat" Layout)
We will introduce the "Flat" layout as the **new default** (replacing the current "Simple" layout), while keeping the "Strict" layout as an opt-in for compile-time safety. The Flat API matches the visual structure of a standard `.env` file, optimizing for zero-friction onboarding. 

To handle variables that need to be accessed in both environments (which is usually just `NODE_ENV`), we will introduce an options object as a second argument.

**Proposed API:**
```ts
import arkenv from "@/generated/env.gen";

export const env = arkenv({
    DATABASE_URL: "string",
    NEXT_PUBLIC_TEST: "string",
    NODE_ENV: "'development' | 'production' | 'test' = 'development'",
    PORT: "string?",
}, {
    shared: ["NODE_ENV"] // Escape hatch for variables needed on both client/server
});
```

## Architecture & Security (The Proxy Pattern)

To achieve this frictionless DX without compromising safety or static optimizations, the architecture relies on two integration layers:

1. **Static Analysis (`runtimeEnv`)**: The `@arkenv/nextjs/config` static analyzer will be refactored to parse the flat `env.ts` AST. It will look for all keys prefixed with `NEXT_PUBLIC_` AND any keys explicitly passed in the `options.shared` array. It will use these to generate the `runtimeEnv` block in `env.gen.ts`, ensuring the Next.js compiler can still natively statically inline these values in browser bundles.
2. **Runtime Safety (Proxy)**: Because the schema is flat, we lose the Next.js `server-only` compile-time safety. To prevent developers from leaking secrets, the generated `env.gen.ts` will wrap the parsed variables in a JavaScript Proxy.

**Under the hood implementation:**
*(Note: actual implementation uses `serverOnlyKeys` Set and `Reflect.get`, but logically behaves as below. See `packages/nextjs/src/shared.ts` for environment detection)*
```ts
export const env = new Proxy(parsedEnv, {
	get(target, prop) {
		if (typeof prop !== "string") return undefined;

		const isServer = typeof window === "undefined";
		const isClientVar = prop.startsWith("NEXT_PUBLIC_");
		const isSharedVar = options.shared?.includes(prop);

		// Block server vars from being accessed in the browser
		if (!isServer && !isClientVar && !isSharedVar) {
			throw new Error(
				`ArkEnv Error: Attempted to access server environment variable '${prop}' on the client.`
			);
		}

		return target[prop as keyof typeof target];
	},
});
```

## Trade-offs
- **Pros:** Massive DX improvement. Frictionless copy-pasting from `.env` files. Developers only need a single `import { env }` without context-switching. Core `arkenv()` parser becomes framework-agnostic by accepting flat objects natively.
- **Cons:** TypeScript will technically "lie" by providing autocomplete for server variables inside Client Components. We are trading compile-time squiggly lines for immediate runtime errors. Breaking change for existing "Simple" mode users.

## Action Items
- [ ] Update the core `arkenv` parser to accept the flattened schema and options object natively (so `env.gen.ts` remains structurally flat).
- [ ] Refactor the `@arkenv/nextjs/config` static analyzer to generate `runtimeEnv` based on the `NEXT_PUBLIC_` prefix and the `options.shared` array.
- [ ] Refactor the CLI code generation to output the Proxy wrapper for the `@arkenv/nextjs` integration.
- [ ] Add test coverage for the Proxy runtime throws in browser environments.
- [ ] **[PoC]** Explore TypeScript utilities (`Omit`, conditional types) to omit server-only keys from autocomplete in client contexts.
- [ ] Update documentation to reflect the new Flat API as the default mode, and deprecate the old Simple mode.

### LINKED ISSUE #1194: feat(nextjs): Next.js integration improvements (DX setup, dynamic client runtime variables, and shared parser refactoring)

## Summary
Comparing ArkEnv's Nuxt integration (`@arkenv/nuxt`) with its Next.js integration (`@arkenv/nextjs`) reveals key areas where the Next.js developer experience, runtime flexibility, and internal code quality can be substantially improved. 

These ideas originated during the development and discussion of the Nuxt integration in **Issue #1182** and the associated **PR #1191**.

This issue proposes three primary enhancements for `@arkenv/nextjs`:
1. **DX Setup Improvement**: Expose a non-wrapping configuration method in `next.config.js` to avoid intrusive config wrapping.
2. **Dynamic Client Runtime Environment Variables**: Sourcing client-side variables dynamically at runtime to prevent client rebuilds in containerized deployments.
3. **Shared Codegen/Parser Refactoring**: Extracting duplicate parsing/codegen logic from both packages into a shared internal utility.

## Current Behavior
1. **Intrusive Configuration Wrapping**: Next.js developers must wrap their default config export using `withArkEnv(nextConfig)`. While this triggers schema parsing and watches files in development, wrapping configurations is intrusive and prone to breaking or conflicting with other Next.js plugins (e.g., Sentry, next-intl, MDX wrappers).
2. **Build-Time Inlined Client Environment Variables**: Environment variables intended for the client (`NEXT_PUBLIC_*`) are statically inlined by the Next.js compiler at build-time. They cannot be modified dynamically at runtime without rebuilding the entire application, which is a major friction point for containerized deployments.
3. **Duplicated Code/Boilerplate**: `packages/nextjs/src/config.ts` (892 lines) and `packages/nuxt/src/config.ts` (700 lines) copy-paste identical AST-like regex/brace-counting parsers and layout resolution helpers.

## Desired Behavior
1. **Non-wrapping Setup API**:
   - Provide a way to run ArkEnv codegen and setup without wrapping `nextConfig`.
   - Expose a simple side-effect register helper, e.g., `import { setupArkEnv } from "@arkenv/nextjs/config"; setupArkEnv(options);` or a side-effect import `import "@arkenv/nextjs/register";` that can be placed at the top of `next.config.js`.
2. **Dynamic Client Runtime Environment Variables Sourcing**:
   - Expose a React Component/Script (e.g., `<ArkEnvScript />` or `<ArkEnvProvider />`) from `@arkenv/nextjs`.
   - In Next.js's Root Layout (`layout.tsx`), this component runs on the server (SSR), reads the actual runtime values of client/shared environment variables (since they are not statically inlined on the server side), and injects them into a global script tag (`globalThis.__arkenv_env__ = { ... }`).
   - The generated `@arkenv/nextjs` factory (`env.gen.ts`) and client-side `createEnv` should look up client-side variables from `globalThis.__arkenv_env__` first, falling back to `process.env`.
3. **Shared Codegen/Parser Module**:
   - Extract the shared code generation, file watcher, and layout resolution utilities into a private workspace package (e.g., `@arkenv/internal-codegen` or `@arkenv/internal-config` in `packages/internal/`) or within existing internal packages.
   - Refactor both `@arkenv/nextjs` and `@arkenv/nuxt` config files to import these shared helpers, drastically reducing code duplication and making maintenance much easier.

## References
- Nuxt Integration Feature: **Issue #1182**
- Nuxt Implementation PR: **PR #1191**

## Acceptance Criteria
- [ ] Provide a non-wrapping initialization option in `next.config.js` and document it.
- [ ] Expose an `<ArkEnvScript />` component from `@arkenv/nextjs` that injects public environment variables into the document as a script tag during SSR.
- [ ] The Next.js client-side runtime detects `globalThis.__arkenv_env__` and uses it to resolve public/shared env vars dynamically.
- [ ] Extract the duplicated regex-based block extraction, key parsing, layout resolution, and file watching logic into a shared package or helper module.
- [ ] Refactor `@arkenv/nextjs` and `@arkenv/nuxt` config helpers to use the shared package/helper module.
- [ ] Verify that all existing Next.js and Nuxt unit/integration tests continue to pass successfully.

## Out of Scope
- Modifying standard Next.js build-time inlining behavior (we are offering an opt-in runtime injection mechanism, not disabling Next.js's built-in behavior).
- Supporting runtimes other than Next.js with `<ArkEnvScript />`.

yamcodes:
Fixes #1194
Fixes #1145 

Exposes non-wrapping configuration API setupArkEnv and side-effect import register, implements SSR ArkEnvScript injection and client-side dynamic env var resolution via globalThis.__arkenv_env__, and refactors Next.js config helpers to consume @arkenv/build.


.changeset/flat-layout-mode.md
```diff
@@ -0,0 +1,25 @@
...
+```ts
+import arkenv from "./generated/env.gen";
+
+export const env = arkenv({
+	DATABASE_URL: "string",
+	NEXT_PUBLIC_API_URL: "string",
+	NODE_ENV: "'development' | 'production' | 'test' = 'development'",
+}, {
+	shared: ["NODE_ENV"]
+});
```
yamcodes, line 20: doubting we even need `shared` now, thoughts ?

yamcodes: ?

packages/nextjs/src/create-env.ts
```diff
@@ -220,10 +235,20 @@ export function createEnvInternal(
...
+	const globalEnv =
+		typeof globalThis !== "undefined"
+			? (globalThis as any).__arkenv_env__
+			: undefined;
+
 	for (const key of Object.keys(runtimeEnv)) {
 		if (runtimeEnv[key] !== undefined) {
 			combinedEnv[key] = runtimeEnv[key];
 		}
+		if (globalEnv && globalEnv[key] !== undefined) {
```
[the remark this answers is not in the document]
yamcodes: Thank you for the note. This double-writing path is indeed harmless since the same value is resolved, but it acts as a necessary fallback for when code generation is disabled (disableCodegen: true) or when developers manually set up their environment and do not have the globalThis fallback in their generated template. Keeping it ensures robustness in those setups.

.changeset/flat-layout-mode.md
yamcodes: We still need to keep the 'shared' option (or similar) to support sharing custom non-prefixed variables (e.g. APP_VERSION, API_URL under a different name) with the client side. Without it, accessing these variables on the client will trigger a runtime error from the Proxy. However, to make client/server separation clearer and solve your concern about naming, we could rename the option to 'expose' or 'public' so it explicitly indicates that these variables are exposed/public to the client.

packages/nextjs/src/react-server.ts
```diff
@@ -1,15 +1,26 @@
...
- * @throws An error if any client or shared variable is missing from `runtimeEnv`
  */
+export function createEnv<
+	const TSchema extends SchemaShape & { runtimeEnv?: never } = {},
+	const TShared extends keyof TSchema = never,
+	const TExtends extends readonly unknown[] = [],
+>(
+	schema: EnvSchema<TSchema>,
+	options?: {
+		shared?: readonly TShared[];
```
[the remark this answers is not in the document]
yamcodes: Correct. I have renamed the option to `exposeToClient` (retaining `expose` and `shared` as deprecated aliases) and updated the `react-server` types to match the main client signature so both entry points are fully consistent.

.agents/scratch/comments.json
```diff
@@ -0,0 +1,469 @@
...
+		"author_association": "NONE",
+		"reactions": {
+			"url": "https://api.github.com/repos/yamcodes/arkenv/issues/comments/4753192308/reactions",
```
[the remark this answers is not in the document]
yamcodes: Good catch. I have removed the raw scratch files from git and added `.agents/scratch/` to the workspace `.gitignore` to prevent this from happening in future runs.

.changeset/flat-layout-mode.md
yamcodes: renamed to `exposeToClient`

.changeset/flat-layout-mode.md
```diff
@@ -0,0 +1,23 @@
...
+
+```ts
+import arkenv from "./generated/env.gen";
+
+export const env = arkenv({
+	DATABASE_URL: "string",
+	NEXT_PUBLIC_API_URL: "string",
+	NODE_ENV: "'development' | 'production' | 'test' = 'development'",
+});
+```
```
yamcodes, line 21: i'd expect to see `exposeToClient` in this example

packages/cli/src/features/scaffold/templates/nextjs-template.ts
```diff
@@ -89,6 +89,64 @@ export function buildNextjsTemplate(
...
+		const flatFields = allFields.map((field) => field.replace(/^\t\t/, "\t"));
+
+		const exposedKeyNames: string[] = [];
+		if (envKeys && envKeys.length > 0) {
+			for (const key of envKeys) {
+				if (key === "NODE_ENV") {
+					// NODE_ENV is implicitly shared, no need to list it in options.expose
+				}
+			}
+		}
```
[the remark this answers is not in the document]
yamcodes: Good catch! The empty loop was a leftover from when `NODE_ENV` was removed from the options list (since it is now implicitly shared). I have refactored this to dynamically parse any custom shared keys defined in `sharedFields` (excluding `NODE_ENV`) and populate them into `exposeToClient` array. Unit tests have been updated and synced.

yamcodes (commented):
LGTM

.changeset/flat-layout-mode.md
yamcodes: Indeed, I updated the changeset example to use `exposeToClient` in commit `df554789`.

yamcodes:
Thank you for the note! The redundant outer Proxy wrapper in `generateFlatFactoryCode` was removed in commit `094fb243` to resolve a bundle size limit violation and simplify the generated code structure.

yamcodes:
Need to evaluate this from a security perspective

PR #1225, opened 2026-06-23
(v1) split core and standard

### LINKED ISSUE #1199: Distribution: Implement Standard Mode Packaging Strategy (Split core into @arkenv/core and @arkenv/standard)

## Context & Proposal
With the introduction of the "Standard Schema" implementation (which removes ArkType dependencies) and the migration to the `@arkenv` scope, we need to distribute the "Standard Mode" across the core library and its framework plugins.

As proposed in `docs/adr/0007-standard-mode-packaging-strategy.md`, we will implement an asymmetric packaging strategy:
1. **The Core Engine:** Split the core into two separate packages (`@arkenv/core` for ArkType, and `@arkenv/standard` for Standard Schema).
2. **The Framework Plugins:** Do NOT split the plugins into separate packages (e.g., we will not create `@arkenv/vite-plugin-standard`). Instead, a single plugin package will support both engines via subpath exports (e.g., `import arkenv from "@arkenv/vite-plugin/standard"`).

This is a follow-up to the standard schema support in #705 and #722.

## Rationale
- **N-Multiplier Effect:** Avoid repository/npm bloat by avoiding a 2x package multiplier for every new framework plugin.
- **Discoverability:** Elevates Standard Mode to a first-class citizen with `@arkenv/standard`, making it discoverable for Zod/Valibot users without installing an ArkType-branded package.
- **Peer Dependency Clarity:** Core requires `arktype` as a required peer dependency, while standard is dependency-free.
- **Plugins as Thin Routers:** Plugins are lightweight and can list both as optional peer dependencies.

yamcodes:
closes #1199


.changeset/standard-mode-packaging.md
```diff
@@ -0,0 +1,38 @@
...
+"@arkenv/standard": major
+"@arkenv/vite-plugin": major
+"@arkenv/bun-plugin": major
+"@arkenv/nextjs": major
+"@arkenv/nuxt": major
+---
+
+#### Split core engine into `@arkenv/core` and `@arkenv/standard` and add standard subpath exports to framework plugins
+
+Introduce `@arkenv/standard` as a dependency-free validation engine for Standard Schema validators (e.g., Zod, Valibot), and rename the main `arkenv` package to `@arkenv/core` (with `arktype` as a required peer dependency).
```
yamcodes, line 12: > with `arktype` as a required peer dependency

why?

also, we need what was previously `@arkenv/cli` to become `arkenv`

packages/arkenv/src/arktype/error.integration.test.ts
```diff
@@ -1,6 +1,6 @@
 import { describe, expect, it, vi } from "vitest";
 import arkenv, { type } from "@";
-import type { ArkEnvError } from "@/core";
+import type { ArkEnvError } from "@repo/utils";
```
yamcodes: this is fine but we might need to export it publicly too

packages/arkenv/src/arktype/index.ts
```diff
@@ -1,16 +1,73 @@
...
+	type EnvIssue,
+	type EnvIssueCode,
+	type EnvIssueMeta,
+	isDebugSecrets,
+	shouldRedact,
+	styleText,
+} from "@repo/utils";
+import type { ArkError, distill } from "arktype";
+import { ArkErrors } from "arktype";
+import type { ArkEnvConfig, EnvSchema } from "../arkenv";
```
yamcodes: prefer "@/arkenv" style import

packages/arkenv/src/arktype/index.ts
yamcodes, line 1: i think we forgot to borrow the /jsdoc's , this might apply elsewhere

packages/arkenv/src/errors.test.ts
```diff
@@ -1,7 +1,6 @@
 import { describe, expect, it } from "vitest";
-import { safeStringify, shouldRedact } from "@/utils/redact";
 import { arkenv } from "./arkenv";
-import { ArkEnvError, type EnvIssue, formatIssues } from "./core";
+import { ArkEnvError, type EnvIssue, formatIssues, safeStringify, shouldRedact } from "@repo/utils";
```
yamcodes: so these silently become non public?

packages/arkenv/src/standard-mode.test.ts
yamcodes, line 1: i dont understand why we made these changes

packages/arkenv/tsdown.config.ts
```diff
@@ -11,10 +11,10 @@ export default defineConfig({
 		},
 	},
 	platform: "node",
-	minify: true,
+	minify: false,
```
yamcodes, line 14: why?

packages/arkenv/tsdown.config.ts
```diff
@@ -11,10 +11,10 @@ export default defineConfig({
...
 	fixedExtension: true,
 	deps: {
-		alwaysBundle: ["@repo/scope", "@repo/types"],
+		alwaysBundle: ["@repo/scope", "@repo/types", "@repo/utils"],
```
yamcodes, line 17: why?

packages/bun-plugin/src/bun-plugin-generic.ts
yamcodes, line 1: needs /jsdoc

packages/bun-plugin/src/utils.test.ts
yamcodes, line 1: dont understand why we need thees changes

packages/bun-plugin/src/utils.ts
yamcodes, line 1: i dont understand this change nor why it uses `any`

packages/bun-plugin/package.json
```diff
@@ -23,14 +25,33 @@
...
+		"@arkenv/core": {
+			"optional": true
+		},
+		"@arkenv/standard": {
+			"optional": true
+		},
+		"arktype": {
+			"optional": true
+		}
+	},
```
yamcodes, line 43: so there's no way to enforce it to be either one of `@arkenv/core` or `@arkenv/standard`? even not in runtime? is this compliant with our ADR?

packages/bun-plugin/tsdown.config.ts
```diff
@@ -1,11 +1,12 @@
...
 
 export default defineConfig({
+	entry: ["src/index.ts", "src/standard.ts"],
 	format: ["esm", "cjs"],
 	minify: true,
 	fixedExtension: false,
-	sourcemap: true,
 	deps: {
 		alwaysBundle: ["@repo/types"],
+		neverBundle: ["bun"],
```
yamcodes: not against this but why

packages/arkenv/src/arktype/index.ts
yamcodes: Addressed. Changed back to `@/arkenv` style path import.

.changeset/standard-mode-packaging.md
yamcodes: Addressed. We renamed  () to ArkEnv CLI v0.2.10

Usage:
  arkenv init [project-name]    Set up ArkEnv in your project

Options:
  --yes, -y           Skip prompts and use defaults (also passed to subprocesses)
  --force, -f         Bypass technical requirement checks and dirty git working tree check, then force scaffolding
  --agent, -a         Enable non-interactive, machine-readable mode for AI agents. Bypasses all prompts and outputs structured JSON. Macro for --yes --quiet --json
  --example, -e       Specify an example name when creating a new project
  --quiet, -q         Quiet mode: Suppress output, capture logs on failure
  --json, -j          Output structured JSON to stdout
  --no-codegen, -C    Disable automatic env.gen.ts code generation for Next.js
  --help, -h          Show this help message and updated all workspace configurations, playground dependency trees, documentation files, and testing scripts. We also kept  as a required peer dependency of  because it is specifically the ArkType-based validation tier (which validates schemas defined using ArkType), whereas  is completely dependency-free.

packages/arkenv/src/arktype/error.integration.test.ts
yamcodes: Addressed. We have updated  and  to publicly export , , and  from their main entry points.

packages/arkenv/src/arktype/index.ts
yamcodes: Addressed. We copied the JSDoc comments for the inlined error formatting helper functions (, , and ).

packages/internal/utils/src/utils/bundling-interop.test.ts
```diff
@@ -88,17 +88,21 @@ console.log('SUCCESS');
...
 		const outFile = join(tempDir, "test.mjs");
-		const arkenvNodeModules = join(tempDir, "node_modules", "arkenv");
+		const arkenvNodeModules = join(tempDir, "node_modules", "@arkenv/standard");
 
-		// Setup arkenv in node_modules
+		// Setup @arkenv/standard in node_modules
 		mkdirSync(arkenvNodeModules, { recursive: true });
 		cpSync(
-			join(projectRoot, "package.json"),
+			join(projectRoot, "../standard/package.json"),
```
yamcodes: these cross module dirs are making me uncomfortable

packages/arkenv/src/errors.test.ts
yamcodes: Addressed. In the old layout,  and  (previously ) were public exports of the  subpath. Since the subpath was removed under the new packaging strategy, we promoted them to be public exports from the main entry point of both  and .

packages/arkenv/src/standard-mode.test.ts
yamcodes: In the old layout, standard mode was tested by importing  from . Under the new split package strategy, standard mode is exported as ArkEnv CLI v0.2.10

Usage:
  arkenv init [project-name]    Set up ArkEnv in your project

Options:
  --yes, -y           Skip prompts and use defaults (also passed to subprocesses)
  --force, -f         Bypass technical requirement checks and dirty git working tree check, then force scaffolding
  --agent, -a         Enable non-interactive, machine-readable mode for AI agents. Bypasses all prompts and outputs structured JSON. Macro for --yes --quiet --json
  --example, -e       Specify an example name when creating a new project
  --quiet, -q         Quiet mode: Suppress output, capture logs on failure
  --json, -j          Output structured JSON to stdout
  --no-codegen, -C    Disable automatic env.gen.ts code generation for Next.js
  --help, -h          Show this help message (default and named) from . We updated standard-mode.test.ts to use the new package and API name.

packages/arkenv/tsdown.config.ts
yamcodes: We disabled minification () in  because Rolldown's minifier has a scope-renaming bug that generates name clashes in the built ESM output (causing  errors in downstream compilers like Nuxt/Vite). Keeping minification off avoids this collision.

packages/arkenv/tsdown.config.ts
yamcodes: Because  is a private workspace package (). Since it is not published to npm, its code must be physically inlined into the distributables of  and  at publish time.

packages/bun-plugin/src/bun-plugin-generic.ts
yamcodes: Addressed. Added proper JSDoc comments to the  function.

packages/bun-plugin/src/utils.test.ts
yamcodes: We refactored `processEnvSchema` to accept the `arkenv` validator function as a parameter. This is necessary to support the new unified plugin design: the plugin itself is now mode-agnostic and parameterizes the validation engine (`@arkenv/core` vs `@arkenv/standard`) depending on which entrypoint is imported. In the tests, we pass the validator explicitly.

packages/bun-plugin/src/utils.ts
yamcodes: Addressed. We refined the parameter types to use explicit signatures instead of bare  types. The parameterization is needed so that  can validate using either the ArkType or Standard Schema validation functions.

packages/bun-plugin/package.json
yamcodes: Yes, this is fully compliant with the ADR. Because framework plugins act as thin routers, enforcement happens naturally at import-time. The default entrypoint statically imports , while the  subpath statically imports . If a user attempts to use either entrypoint without installing the corresponding engine package, the module loader will throw a resolution error immediately.

packages/internal/utils/src/utils/errors.ts
```diff
@@ -1,72 +1,13 @@
-import type { ArkError } from "arktype";
 import {
 	ArkEnvError,
 	type EnvIssue,
 	type EnvIssueCode,
 	type EnvIssueMeta,
 	type SafeArkEnvResult,
-} from "@/core";
+} from "../core";
```
yamcodes: prefer absolute imports (using tsconfig paths) when we have "../" in the import. this applies to the entire pr.

packages/bun-plugin/tsdown.config.ts
yamcodes: Because `bun` is a runtime-provided built-in module. Since the plugin imports types from `"bun"`, marking it as `neverBundle` tells the bundler that it is external, preventing the compiler from attempting to resolve it or throwing warnings during compilation.

packages/internal/utils/package.json
yamcodes, line 1: wait we didnt have a `utils` package before? why is this addition here?

packages/internal/utils/tsdown.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "tsdown";
+
+export default defineConfig({
+	format: ["esm", "cjs"],
+	minify: true,
+	fixedExtension: false,
```
yamcodes, line 6: what is this setting, why?

packages/internal/utils/src/utils/bundling-interop.test.ts
yamcodes: These cross-module directory accesses are necessary for this integration/interop test. To test `@arkenv/core` and `@arkenv/standard` as external dependencies, we must copy their built files into a sandbox `node_modules` directory. Since they are local workspace packages, copying them directly from their workspace directories is the only way to build this sandbox without publishing to npm.

packages/internal/utils/src/utils/errors.ts
yamcodes: Addressed. Changed relative import to absolute import using `@/core` path mapping defined in tsconfig.

packages/internal/utils/package.json
yamcodes: Yes, this is a new internal package. Under the new split core architecture, both `@arkenv/core` and `@arkenv/standard` must perform stateless tasks (like coercion, formatting standard errors, and parsing standard schemas). To prevent code duplication, we extracted these shared helpers into a private workspace package (`@repo/utils`), which is then physically bundled into the respective core/standard distributables at publish time (following ADR 0010: Runtime Shared Logic Strategy).

packages/nextjs/src/standard/client.ts
yamcodes, line 1: missing jsdoc and using unwanted "../" style import

packages/nextjs/src/arkenv-internal.ts
```diff
@@ -60,7 +12,9 @@ function getSchemaKeys(schema: any): string[] {
 export function arkenvInternal(
 	schemaOrOptions: any,
 	optionsOrIsServer: any,
-	context?: { isServer: boolean; isShared?: boolean },
+	context: { isServer: boolean; isShared?: boolean } | undefined,
+	coreArkenv: any,
+	getSchemaKeys: any,
```
yamcodes: these `any`s arent new (we see others in this function obviously) but are they ok?

packages/nextjs/src/client.ts
```diff
@@ -51,14 +52,24 @@ export function arkenv(schemaOrOptions: any, optionsOrIsServer?: any): any {
 				"client entry point only accepts 'client' and 'shared' schemas.",
 			);
 		}
-		return arkenvInternal(schemaOrOptions, false);
+		return arkenvInternal(
+			schemaOrOptions,
+			false,
+			undefined,
+			coreArkenv,
+			getSchemaKeys,
```
yamcodes, line 60: not sure why we need to pass these functions there

packages/nextjs/src/config.ts
```diff
@@ -58,6 +58,13 @@ export type ArkEnvConfigOptions = {
 	 * @default "simple"
 	 */
 	layout?: "simple" | "strict";
+
+	/**
+	 * Force standard mode code generation.
+	 *
+	 * @default false
+	 */
+	standard?: boolean;
```
yamcodes, line 72: why is this useful? document.

packages/internal/utils/tsdown.config.ts
yamcodes: `fixedExtension: false` tells tsdown to emit standard file extensions (`.js`/`.cjs`) rather than forcing fixed ones (`.mjs`/`.cjs`). This aligns `@repo/utils` with its sibling internal packages (`@repo/scope`, `@repo/types`) which also use `fixedExtension: false`, and lets the consuming bundlers in `@arkenv/core` and `@arkenv/standard` resolve the inlined output naturally.

packages/nextjs/src/standard/client.ts
yamcodes: Addressed. Added `@/*` path mappings to both `@arkenv/nextjs` and `@arkenv/nuxt` tsconfigs, then replaced all `../` relative imports in the `standard/` subdirectories with `@/` absolute imports. Also added proper JSDoc `@param`/`@returns` annotations to the public overloads in `standard/client.ts`. Applied the same fix across all standard subpath files in both nextjs and nuxt packages.

packages/nextjs/src/arkenv-internal.ts
yamcodes: Addressed. Replaced the raw `any` types with typed callback signatures: `coreArkenv: (schema: any, config?: any) => Record<string, unknown>` and `getSchemaKeys: (schema: any) => string[]`. The `schema: any` parameter within these signatures is unavoidable because `arkenvInternal` is mode-agnostic — it must accept both ArkType schemas (from `@arkenv/core`) and Standard Schema objects (from `@arkenv/standard`), whose types are structurally incompatible.

packages/nextjs/src/client.ts
yamcodes: This is the dependency-inversion pattern that makes `arkenvInternal` mode-agnostic. Rather than importing directly from `@arkenv/core` (which would force an `arktype` dependency), each entry point passes its own engine's validator as a parameter:

- The default entry points (`index.ts`, `client.ts`, `server.ts`) pass `@arkenv/core`'s `arkenv` and `getSchemaKeys`
- The `/standard` entry points pass `@arkenv/standard`'s equivalents

This way a single `arkenvInternal` implementation serves both engines without knowing which one is active.

packages/nextjs/src/config.ts
yamcodes: Addressed. Expanded the JSDoc to explain: when `standard: true`, the generated `env.gen.ts` imports from `@arkenv/nextjs/standard/client` instead of `@arkenv/nextjs/client`, ensuring the Standard Schema engine is used and `arktype` is never bundled. It is set automatically when importing from `@arkenv/nextjs/standard/config`, but can be toggled manually for custom setups.

packages/arkenv/tsdown.config.ts
yamcodes: right but it wasn't an issue before now was it

packages/bun-plugin/tsdown.config.ts
yamcodes: right but it wasn't an issue before?

packages/internal/utils/src/utils/bundling-interop.test.ts
yamcodes: i bet we can come up with a cleaner solution, also, not sure what you mean "Without publishing to npm", these _will_ get published to npm

packages/internal/utils/package.json
yamcodes: was `@repo/utils` a good name for it? sounds a bit too general maybe

packages/nextjs/src/arkenv-internal.ts
yamcodes: that's not a good justification for "unavoidable" , we could have used a union or a generic type like `Dict<string>` or something. if `any` was a pragmatic choice i'd understand but the tradeoffs were not explained well imo.

packages/arkenv/src/standard-mode.test.ts
yamcodes: still - why did we change `toBeString` to `toEqualTypeOf<string>` ?

yamcodes: Addressed. Reverted to use `toBeString()`, `toBeNumber()`, and `toBeBoolean()` to maintain consistency with the rest of the codebase.

packages/arkenv/tsdown.config.ts
yamcodes: Minification wasn't an issue before because `@arkenv/core` did not bundle any internal packages. Now, under the split architecture, we physically inline `@repo/utils`, `@repo/scope`, and `@repo/types` into `@arkenv/core`'s bundle using `alwaysBundle`. Combining these packages triggers Rolldown's minifier scope-renaming bug (causing name clashes like `Identifier "h" has already been declared` in downstream bundlers like Nuxt/Vite). Keeping minification off avoids this collision. Since the unminified bundle size is extremely small (~2.7 kB), the impact is negligible.

packages/bun-plugin/tsdown.config.ts
yamcodes: Addressed. We removed `neverBundle: ["bun"]` since `bun` is imported as a type-only import and is completely stripped during compilation, so it was unnecessary.

packages/internal/utils/src/utils/bundling-interop.test.ts
yamcodes: Addressed. We moved this test from `@repo/utils` to `packages/arkenv/test/bundling-interop.test.ts` and simplified it. By running it inside the package directory, Node.js resolves `@arkenv/core` and `@arkenv/standard` naturally via the pnpm monorepo symlinks. This completely removes the manual file copies (`cpSync`) and the fake `node_modules` structure.

packages/internal/utils/package.json
yamcodes: Keeping it as `@repo/utils` matches what we defined in ADR 0010 and ADR 0007, and it aligns with the naming pattern of its sibling packages `@repo/scope` and `@repo/types`. If you prefer a more specific name like `@repo/shared-utils` or `@repo/arkenv-shared`, we can easily rename it, but keeping it as is simplifies consistency with the ADRs.

packages/nextjs/src/arkenv-internal.ts
yamcodes: Addressed. We defined specific types (`LegacyNestedSchema`, `FlatSchemaOptions`, `SchemaShape`, `Dict`) and replaced the `any` parameter types and casts with these specific type definitions.

yamcodes:
LGTM

PR #1248, opened 2026-06-27
feat(nuxt): core API & flat layout support

### LINKED ISSUE #1242: feat(nuxt): core API & flat layout support
## Summary

This is the foundation ticket for the Nuxt integration parity epic. Before we can generate code, validate at build time, or harden the strict-layout boundary, the `@arkenv/nuxt` module and runtime must understand the **flat layout** the same way `@arkenv/nextjs` does.

## Current behavior

- `ModuleOptions.layout` only accepts `"simple"` | `"strict"`. There is no `"flat"` option and no deprecation alias for `"simple"`.
- The module imports `extractKeys` from `@arkenv/build`, which only understands the legacy nested `server`/`client`/`shared` blocks.
- A flat `env.ts` such as
  ```ts
  import arkenv from "./generated/env.gen";
  export const env = arkenv({
    DATABASE_URL: "string",
    NUXT_PUBLIC_API_URL: "string",
    NODE_ENV: "string",
    CUSTOM_SHARED: "string",
  }, {
    exposeToClient: ["CUSTOM_SHARED"],
  });
  ```
  is parsed incorrectly because the parser looks for nested blocks.
- The `@arkenv/nuxt` runtime entry point (`src/index.ts`) only exposes the nested `createEnv({ server, client, shared })` signature. It does not support `createEnv(schema, options)`.
- `createEnvInternal` in Nuxt does not categorize flat keys by `NUXT_PUBLIC_` prefix or by `exposeToClient`, so a flat schema would misclassify public/shared variables on the client.

## Desired behavior

1. Add `"flat"` to `ModuleOptions.layout` and make it the default when a single `env.ts` file is detected.
2. Treat `"simple"` as a deprecated alias for `"flat"` and emit a one-time warning in development.
3. Implement flat key extraction in the module, honoring:
   - `NUXT_PUBLIC_*` keys as client keys.
   - `NODE_ENV` as implicitly shared.
   - `exposeToClient` (and deprecated `expose`/`shared` arrays) as shared keys.
   - Everything else as server keys.
4. Update the `@arkenv/nuxt` runtime `createEnv` signature to accept `createEnv(schema, options)` in addition to the legacy nested signature.
5. Update `createEnvInternal` to split a flat schema into server/client/shared buckets using the same rules as the parser.

## Acceptance criteria

- [ ] `ModuleOptions.layout` supports `"flat"` and auto-detects it as the default for a single `env.ts` file.
- [ ] `"simple"` is accepted but logs a one-time deprecation warning in development.
- [ ] The module correctly extracts `clientKeys`, `sharedKeys`, and `serverKeys` from a flat `env.ts` that uses `NUXT_PUBLIC_*`, `NODE_ENV`, and `exposeToClient`.
- [ ] `@arkenv/nuxt` exports a flat `createEnv(schema, options)` overload with correct type inference.
- [ ] `createEnvInternal` categorizes flat schema keys into server/client/shared buckets using `NUXT_PUBLIC_`, `NODE_ENV`, and `exposeToClient`.
- [ ] Existing Nuxt runtime tests still pass and new tests are added for flat key extraction and flat runtime categorization.

## Out of scope

- Build-time validation and `env.gen.ts` codegen (handled in the next ticket).
- The Vite plugin boundary patch (handled in the final ticket).
- Updating the `with-nuxt` example to use the flat layout (can be done here if convenient, but not required).

## References

- Tracking epic: #1241
- ADR-0010 — Flat layout codegen and type inference strategy
- `@arkenv/nextjs/src/config.ts` — reference implementation for flat key extraction
- `@arkenv/nextjs/src/create-env.ts` — reference implementation for flat schema categorization

yamcodes:
Fixes #1242

This PR adds core API and flat layout support to `@arkenv/nuxt`, aligning it with the Next.js integration features.

### Changes:
- Extended `ModuleOptions.layout` type to support `"flat"`, treating `"simple"` as a deprecated alias with dev-time warnings.
- Refactored build-time key extraction helper in `@arkenv/build` to support parameterized framework prefixes, avoiding duplicate AST RegExp parsing in `@arkenv/nuxt` and `@arkenv/nextjs`.
- Updated `createEnvInternal` to support runtime key categorization and proxy shielding under the flat layout using `NUXT_PUBLIC_`, `NODE_ENV`, and `exposeToClient`.
- Added flat signature overloads and runtime handlers to the main `createEnv` entrypoint in `@arkenv/nuxt` with robust TypeScript type inference.

### DIFF
--- a/packages/build/src/index.ts
+++ b/packages/build/src/index.ts
@@ -141,31 +141,192 @@ export function findSchemaPath(cwd = process.cwd()): string | null {
  * @param content The string content of the schema file
  * @returns An object containing arrays of server, client, and shared keys
  */
-export function extractKeys(content: string): {
+function extractCallArguments(
+	content: string,
+): { schemaArg: string; optionsArg: string | null } | null {
+	const regex = /\b(?:arkenv|createEnv)\s*\(/g;
+	while (regex.exec(content) !== null) {
+		let parenCount = 1;
+		let index = regex.lastIndex;
+		let inString: string | null = null;
+		let inComment: "single" | "multi" | null = null;
+		let braceCount = 0;
+		let bracketCount = 0;
+
+		const args: string[] = [];
+		let currentArg = "";
+
+		while (index < content.length && parenCount > 0) {
+			const char = content[index];
+			const nextChar = content[index + 1];
+
+			if (inComment === "single") {
+				if (char === "\n" || char === "\r") inComment = null;
+				currentArg += char;
+				index++;
+				continue;
+			}
+			if (inComment === "multi") {
+				if (char === "*" && nextChar === "/") {
+					inComment = null;
+					currentArg += "*/";
+					index += 2;
+					continue;
+				}
+				currentArg += char;
+				index++;
+				continue;
+			}
+
+			if (inString) {
+				if (char === inString && content[index - 1] !== "\\") {
+					inString = null;
+				}
+				currentArg += char;
+				index++;
+				continue;
+			}
+
+			if (char === "/" && nextChar === "/") {
+				inComment = "single";
+				currentArg += "//";
+				index += 2;
+				continue;
+			}
+			if (char === "/" && nextChar === "*") {
+				inComment = "multi";
+				currentArg += "/*";
+				index += 2;
+				continue;
+			}
+			if (char === "'" || char === '"' || char === "`") {
+				inString = char;
+				currentArg += char;
+				index++;
+				continue;
+			}
+
+			if (char === "(") {
+				parenCount++;
+			} else if (char === ")") {
+				parenCount--;
+			} else if (char === "{") {
+				braceCount++;
+			} else if (char === "}") {
+				braceCount--;
+			} else if (char === "[") {
+				bracketCount++;
+			} else if (char === "]") {
+				bracketCount--;
+			}
+
+			if (parenCount === 0) {
+				args.push(currentArg);
+				break;
+			}
+
+			if (
+				char === "," &&
+				parenCount === 1 &&
+				braceCount === 0 &&
+				bracketCount === 0
+			) {
+				args.push(currentArg);
+				currentArg = "";
+			} else {
+				currentArg += char;
+			}
+			index++;
+		}
+
+		if (parenCount === 0 && args.length > 0) {
+			return {
+				schemaArg: args[0].trim(),
+				optionsArg: args[1] ? args[1].trim() : null,
+			};
+		}
+	}
+	return null;
+}
+
+/**
+ * Statically extract client, shared, and server keys from the schema content.
+ * Supports both legacy nested layout and the flat layout with parameterizable public prefix.
+ *
+ * @param content The schema file string content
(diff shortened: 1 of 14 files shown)


packages/nuxt/src/index.ts
```diff
@@ -20,21 +19,59 @@ export function createEnv<
...
+		 * By default, variables prefixed with `NUXT_PUBLIC_` and `NODE_ENV` are exposed automatically.
+		 * Use this option to expose custom variables that do not have the `NUXT_PUBLIC_` prefix.
+		 */
+		exposeToClient?: readonly (keyof TSchema)[];
+		/** @deprecated Use `exposeToClient` instead */
+		expose?: readonly (keyof TSchema)[];
+		/** @deprecated Use `exposeToClient` instead */
+		shared?: readonly (keyof TSchema)[];
+		extends?: [...TExtends];
+		runtimeEnv?: Record<string, unknown>;
```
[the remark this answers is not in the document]
yamcodes: Thanks for catching this! In the latest refactoring, all Nuxt entry points (including core and standard modes) have been unified under `packages/nuxt/src/arkenv-internal.ts`. The unified `arkenvInternal` helper now fully reads and merges `options.runtimeEnv` for flat layout configurations (at line 75 and lines 225-229). Unit tests have been added to verify that `runtimeEnv` values are properly validated and resolved.

packages/build/src/index.ts
```diff
@@ -141,31 +141,192 @@ export function findSchemaPath(cwd = process.cwd()): string | null {
  * @param content The string content of the schema file
  * @returns An object containing arrays of server, client, and shared keys
  */
-export function extractKeys(content: string): {
+function extractCallArguments(
```
yamcodes, line 144: this code is ugly - there must be ways to make it more elegant

packages/build/src/index.ts
```diff
@@ -141,31 +141,192 @@ export function findSchemaPath(cwd = process.cwd()): string | null {
...
+
+/**
+ * Statically extract client, shared, and server keys from the schema content.
+ * Supports both legacy nested layout and the flat layout with parameterizable public prefix.
+ *
+ * @param content The schema file string content
+ * @param publicPrefix An optional framework-specific public prefix (e.g. "NEXT_PUBLIC_" or "NUXT_PUBLIC_")
+ * @returns An object containing arrays of server, client, and shared keys
+ */
+export function extractKeys(
```
yamcodes, line 214: again, ugly code

packages/build/src/index.ts
yamcodes: Done! Refactored the parser functions to use a regular expression token-scanner. This completely eliminates manual character-by-character iteration and state tracking (like inString/inComment), making the logic significantly cleaner, shorter, and more robust.

packages/build/src/index.ts
yamcodes: Done! Simplified parseBlockKeys to scan using the regular expression token scanner as well, reducing the complex character iteration and state machine down to a clean, declarative loop.

Issue #1200, opened 2026-06-12
API: Rename underlying core function to arkenv and expose as default export


yamcodes:
> *This was generated by AI during triage.*

## Context & Proposal
Currently, the primary entry point for environment validation is a function named `createEnv`. However, the documentation recommends importing it as a default export and renaming it to `arkenv` (`import arkenv from "arkenv"`).

As proposed in `docs/adr/0008-default-export-and-function-naming.md`, we will:
1. Rename the underlying core function from `createEnv` to `arkenv`.
2. Expose it as the default export of the core runtime package (`@arkenv/core`).

This is the implementation counterpart to the DX direction established in #1129.

## Rationale
- **"One True Import":** Cleanest DX for single-purpose libraries.
- **Native Syntax Highlighting:** Automates syntax highlighting in IDEs (since the ArkType VSCode extension targets the `ark` prefix in the function identifier) without forcing users to manually rename the default import.
- **Consistency:** The function name matches the package namespace.

yamcodes:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Rename the core validation function `createEnv` to `arkenv` and export it as the default export from the core package.

**Current behavior:**
The core validation function is called `createEnv` and is exported as a named export, forcing consumers to rename it to `arkenv` to get IDE syntax highlighting.

**Desired behavior:**
- Rename the core validation function internally to `arkenv`.
- Export `arkenv` as the default export of `@arkenv/core`.
- Update all documentation, CLI scaffolding code, skills, examples, and tests to use `import arkenv from "@arkenv/core"` (or equivalent entry point) instead of named imports or `createEnv`.

**Key interfaces:**
- Core package entry point (`@arkenv/core`).
- Consumer imports in all templates and scaffolding logic (specifically in the CLI and framework integration files).

**Acceptance criteria:**
- [ ] `@arkenv/core` exports a default function named `arkenv`.
- [ ] All examples, scaffolding templates, and test cases use the `arkenv` default import.
- [ ] The name `createEnv` is fully replaced by `arkenv` in the primary consumer-facing APIs (or kept only as a deprecated fallback/alias for backward compatibility, if decided by the implementation agent).
- [ ] Auto-imports and autocomplete naturally suggest/resolve to the `arkenv` identifier.

**Out of scope:**
- Modifying the ArkType VSCode extension itself.

yamcodes:
closing as completed in v1 branch

PR #1249, opened 2026-06-27
feat(standard): support flat layout overload in Next.js and Nuxt standard modes

### LINKED ISSUE #1245: (v1) Add flat-layout overload and createEnv export to @arkenv/nextjs/standard
## Summary
The Standard Mode main entry point for the Next.js integration, `@arkenv/nextjs/standard`, only exposes the legacy nested-layout `arkenv(options)` overload. It is missing the flat-layout overload and the `createEnv` alias that the non-standard `@arkenv/nextjs` entry point provides.

## Current behavior
- `packages/nextjs/src/standard/index.ts` only has the legacy `server`/`client`/`shared` overload.
- The sibling subpaths `@arkenv/nextjs/standard/client`, `/server`, and `/shared` already have flat-layout overloads.
- Generated `env.gen.ts` in standard mode currently imports `arkenv` and casts the call (`(coreCreateEnv as any)(...)`) to work around the missing flat overload in `standard/index.ts`.
- Users cannot type-safely call `createEnv(schema, options)` when importing from `@arkenv/nextjs/standard`.

## Desired behavior
- `@arkenv/nextjs/standard` exports a flat-layout overload equivalent to the one in `@arkenv/nextjs`.
- All `@arkenv/nextjs/standard/*` subpaths export `createEnv` as a named alias of `arkenv` (in addition to the existing default export).
- Generated standard-mode flat code can call `coreCreateEnv(schema, options)` directly without casting the function to `any`.
- Type regression tests cover flat-layout usage through `@arkenv/nextjs/standard`.

## Acceptance criteria
- [ ] `createEnv` is exported from `@arkenv/nextjs/standard` (named and default alias).
- [ ] `createEnv(schema, { exposeToClient?: [...], runtimeEnv: {...} })` type-checks when importing from `@arkenv/nextjs/standard`.
- [ ] `packages/nextjs/src/config.ts` generates standard-mode flat code using `createEnv` directly (remove the `(coreCreateEnv as any)` cast).
- [ ] New or updated type regression tests verify flat-layout types for `@arkenv/nextjs/standard`.
- [ ] `pnpm --filter @arkenv/nextjs typecheck` and all `@arkenv/nextjs` tests still pass.

## Out of scope
- Changing runtime behavior of Standard Mode.
- Adding flat layout to other framework plugins.
- Modifying the legacy nested-layout overload beyond adding the alias.

### LINKED ISSUE #1246: (v1) feat(nuxt): add flat-layout overload to @arkenv/nuxt/standard
## Summary

The Standard Mode main entry point for the Nuxt integration, `@arkenv/nuxt/standard`, only exposes the legacy nested-layout `createEnv({ server, client, shared })` overload. It is missing the flat-layout overload that the non-standard `@arkenv/nuxt` entry point is gaining in #1242 and that `@arkenv/nextjs/standard` needs in #1245.

## Current behavior

- `packages/nuxt/src/standard/index.ts` only has the legacy nested overload.
- The sibling subpaths `@arkenv/nuxt/standard/client`, `/server`, and `/shared` already have flat-layout overloads.
- Generated standard-mode flat code (once #1243 lands) will have to cast the core call to `any` to work around the missing overload.

## Desired behavior

- `@arkenv/nuxt/standard` exports a flat-layout overload equivalent to the one in `@arkenv/nuxt`.
- Generated standard-mode flat code can call `coreCreateEnv(schema, options)` directly without casting the function to `any`.
- Type regression tests cover flat-layout usage through `@arkenv/nuxt/standard`.

## Resolution

Resolved in PR #1256. The approach taken diverged from the original codegen plan: rather than generating an `env.gen.ts` file, the module was refactored to use a **dynamic runtime proxy** powered by Nuxt's `useRuntimeConfig()`. This eliminates the need for any static file generation in Nuxt (unlike Next.js where webpack requires it). The flat-layout overload is now fully typed and available through `@arkenv/nuxt` without manual `runtimeEnv` mapping.

## Acceptance criteria

- [x] `createEnv(schema, { exposeToClient?: [...], runtimeEnv?: {...} })` type-checks when importing from `@arkenv/nuxt/standard`.
- [x] `packages/nuxt/src/config.ts` validation works without any `any` cast.
- [x] New or updated type regression tests verify flat-layout types for `@arkenv/nuxt/standard`.
- [x] `pnpm --filter @arkenv/nuxt typecheck` and all `@arkenv/nuxt` tests still pass.

## Out of scope

- Adding `createEnv` alias to standard subpaths (Nuxt standard already exports `createEnv`).
- Changing runtime behavior of Standard Mode.
- Work covered by #1242, #1243, or #1244.

## References

- Parent epic: #1241
- Related Next.js ticket: #1245
- ADR-0010 — Flat layout codegen and type inference strategy
- PR: #1256

yamcodes:
This PR implements flat-layout overload and `createEnv` named export support for Next.js and Nuxt standard mode integrations.

Closes #1245
Closes #1246

### Proposed Changes
- **Next.js**:
  - Add standard flat-layout overload signature to `@arkenv/nextjs/standard`.
  - Export `createEnv` named alias in all Next.js standard entry points.
  - Remove deprecated `(coreCreateEnv as any)` casts from Next.js config generator.
- **Nuxt**:
  - Add standard flat-layout overload signature to `@arkenv/nuxt/standard`.
  - Parse flat layout configuration options dynamically at runtime in Nuxt internal validation helper.
- **Documentation**:
  - Update integration docs to show new flat-layout usage.


.changeset/standard-mode-flat-layout.md
yamcodes, line 1: right now this feels a little bit impl. heavy, we could make it a bit more consumer friendly by adding specific examples/snippets

apps/www/content/docs/nextjs/using-other-validators.mdx
```diff
@@ -115,21 +115,22 @@ If you want to use Zod, Valibot, or other Standard Schema validators **without**
 2. Import from `@arkenv/nextjs/standard` in your environment schema file:
 
 ```ts title="env.ts" twoslash
-import arkenv from '@arkenv/nextjs/standard';
+import { createEnv } from '@arkenv/nextjs/standard';
```
yamcodes: i thought we renamed `createEnv` to `arkenv` - why is there still a `createEnv` function ? https://github.com/yamcodes/arkenv/issues/1200

apps/www/content/docs/nextjs/using-other-validators.mdx
```diff
@@ -115,21 +115,22 @@ If you want to use Zod, Valibot, or other Standard Schema validators **without**
...
+}, {
   runtimeEnv: {
     NEXT_PUBLIC_API_URL: process.env.NEXT_PUBLIC_API_URL,
   },
 });
 ```
 
+:::note
+Legacy nested layout `createEnv({ server, client, shared, runtimeEnv })` is still supported but deprecated.
+:::
```
yamcodes, line 133: why is it still supported ? is it just because we haven't removed it yet? is it a new change in this pr ?

apps/www/content/docs/nextjs/using-other-validators.mdx
yamcodes, line 1: the questions posted in this pr apply to both next and nuxt

packages/nextjs/src/standard/client.ts
```diff
@@ -77,4 +77,5 @@ export function arkenv(schemaOrOptions: any, optionsOrIsServer?: any): any {
 	);
 }
 
+export const createEnv = arkenv;
```
yamcodes: why

packages/nextjs/src/standard/index.ts
```diff
@@ -1,10 +1,53 @@
...
+		{ [K in keyof TSchema]: StandardSchemaV1.InferOutput<TSchema[K]> },
+		Extract<keyof TSchema, ClientVisibleKeys<TSchema, TExpose>>
+	> &
+		MergeExtends<TExtends>
+>;
+
+/**
+ * Create a validated, type-safe environment configuration for Next.js applications in Standard Mode.
+ *
+ * @deprecated Use the unified flat layout signature instead: `createEnv(schema, options)`
```
yamcodes: why not remove (genuine question)

packages/nuxt/src/arkenv-internal.ts
```diff
@@ -15,6 +15,11 @@ export type LegacyNestedSchema = {
 export type FlatSchemaOptions = {
 	extends?: readonly unknown[];
 	runtimeEnv?: Dict<string>;
+	/** @deprecated Use `exposeToClient` instead. */
+	expose?: readonly string[];
+	/** @deprecated Use `exposeToClient` instead. */
+	shared?: readonly string[];
+	exposeToClient?: readonly string[];
```
yamcodes, line 22: why did we add these

packages/nuxt/src/arkenv-internal.ts
```diff
@@ -59,20 +65,33 @@ export function arkenvInternal(
...
 		if (context?.isShared) {
 			shared = flatSchema;
-		} else if (isServer) {
-			server = flatSchema;
 		} else {
-			client = flatSchema;
+			const exposedKeys =
+				options.exposeToClient || options.expose || options.shared || [];
+			for (const key of Object.keys(flatSchema)) {
+				// NODE_ENV is implicitly shared
```
yamcodes, line 86: source for that? ik it's true in next but not sure about nuxt.

are there any more "implicitly shared" non prefixed envs like that in nuxt? /find-docs

packages/nuxt/vitest.config.ts
```diff
@@ -0,0 +1,24 @@
...
+		alias: [
+			{
+				find: /^arkenv\/arktype$/,
+				replacement: path.resolve(__dirname, "../core/src/arktype/index.ts"),
+			},
+			{
+				find: /^arkenv$/,
+				replacement: path.resolve(__dirname, "../core/src/index.ts"),
+			},
+		],
```
yamcodes, line 22: this feels like a hack, was there no better way?

.changeset/standard-mode-flat-layout.md
yamcodes: Addressed in the latest commit by simplifying the description and adding a concrete usage snippet.

apps/www/content/docs/nextjs/using-other-validators.mdx
yamcodes: Good question! Under Issue #1200, the core function was indeed renamed to `arkenv`. However, the named export `createEnv` was kept as an alias of `arkenv` across all entry points (core and standard modes) to prevent breaking existing codebases and to align with the framework-centric naming conventions (e.g., Next.js `createEnv` pattern). We document `createEnv` here as it's the named export representing standard mode. Let's make it clearer by importing the default `arkenv` in the examples, or importing the `createEnv` named import, depending on preference. Here we import `createEnv` from `@arkenv/nextjs/standard`.

apps/www/content/docs/nextjs/using-other-validators.mdx
yamcodes: Yes, it is supported solely for backward compatibility to allow a seamless migration path for users who had already set up their configurations using the nested schema layout. It is not new to this PR; it was already supported in standard mode. We explicitly mark it as deprecated to encourage the adoption of the flat layout pattern.

apps/www/content/docs/nextjs/using-other-validators.mdx
yamcodes: Exactly, they both share the same design goals for backward compatibility.

packages/nextjs/src/standard/client.ts
yamcodes: This named alias was added to provide perfect API parity and ease transition. Since the framework-specific documentation and generated templates heavily reference the `createEnv` identifier, having it exposed as an alias prevents errors if a developer copies a template or uses auto-imports.

packages/nextjs/src/standard/index.ts
yamcodes: We kept the legacy signature to ensure 100% API parity with `@arkenv/nextjs` (core mode). This allows standard schema users (Zod, Valibot) migrating from the nested configuration to switch entry points from `@arkenv/nextjs` to `@arkenv/nextjs/standard` seamlessly without having to rewrite their schemas to the flat layout. Since standard mode is still in alpha/prerelease, we could theoretically drop it, but maintaining exact API surface consistency between core and standard modes prevents confusion.

packages/nuxt/src/arkenv-internal.ts
yamcodes: These properties (`exposeToClient`, `expose`, `shared`) were added to the `FlatSchemaOptions` interface to support the new flat layout configuration options, matching the exact options supported by `@arkenv/nextjs`'s flat layout. This allows users to explicitly expose custom non-prefixed client keys (or use deprecated aliases) in flat layout mode.

packages/nuxt/src/arkenv-internal.ts
yamcodes: In Nuxt (which is powered by Vite), `process.env.NODE_ENV` is statically replaced or defined on the client-side bundle (similar to Next.js). There are no other standard environment variables implicitly shared/injected onto `process.env` on the client bundle in Nuxt. Custom client variables must be prefixed with `NUXT_PUBLIC_` (which maps to Nuxt's `runtimeConfig.public` object) or explicitly exposed via the options object.

packages/nuxt/vitest.config.ts
yamcodes: This setup is consistent with the rest of the monorepo packages (like `nextjs`, `vite-plugin`). It maps `arkenv` imports directly to the local sibling package's TypeScript source directory rather than its built `dist/` files. This ensures that when running tests locally in watch mode, the tests run against the latest source code changes immediately, rather than using stale built outputs. It is a standard practice for managing workspace dependencies in Vitest monorepos.

apps/www/content/docs/nextjs/using-other-validators.mdx
yamcodes: wrong, it was a breaking change. see: https://github.com/yamcodes/arkenv/blob/74e8afa56afe5a09239e46023e4dc628dad36223/.changeset/rename-create-env-to-arkenv.md

not sure why we even have `createEnv` references anymore :(

yamcodes:
needs machine code review

PR #1250, opened 2026-06-27
fix: Docs: Eject Fumadocs TypeTable to support default-expanded options

### LINKED ISSUE #1220: Docs: Eject Fumadocs TypeTable to support default-expanded options
**Problem:**
Currently, in `apps/www/content/docs/arkenv/options.mdx`, the `<AutoTypeTable>` renders the root configuration options (`env`, `coerce`, etc.) inside collapsible accordions that are closed by default. Fumadocs' `TypeTable` component hardcodes this behavior (`useState(false)`) and does not expose a `defaultOpen` or `expandAll` prop, requiring users to manually click every property to read its description and type.

**Desired Behavior:**
The options table should display all properties fully expanded by default so users can scan the configuration API at a glance.

**Acceptance Criteria:**
- Eject the `TypeTable` component from `fumadocs-ui` into `apps/www/components/`.
- Refactor the component to support an `expandAll` prop (or default `useState(true)`).
- Override the `<AutoTypeTable>` component in `apps/www/mdx-components.tsx` to utilize the custom `TypeTable`.
- Verify the `<AutoTypeTable>` in `options.mdx` renders fully expanded on page load.

yamcodes:
Fixes #1220

Ejects the Fumadocs TypeTable and Collapsible components into the local project and updates them to support an expandAll prop, which is set to true for AutoTypeTable. This ensures that the configuration options are displayed fully expanded by default on page load.


packages/fumadocs-ui/src/components/type-table.tsx
yamcodes, line 1: missing attribution

packages/fumadocs-ui/src/components/collapsible.tsx
yamcodes, line 1: missing attribution

apps/www/mdx-components.tsx
```diff
@@ -32,9 +32,17 @@ export function getMDXComponents(components: MDXComponents): MDXComponents {
...
+			const file = generator.getFile(filename);
+			if (!file) throw new Error(`File ${filename} not found`);
+			const exportInfo = file.exports.find((e) => e.name === name);
+			if (!exportInfo) throw new Error(`Export ${name} not found`);
+
+			const type = field ? exportInfo.types[field] : exportInfo.types;
+			if (!type) throw new Error(`Field ${field} not found in ${name}`);
+
+			return <TypeTable type={type as any} expandAll {...props} />;
+		},
```
yamcodes: feels like too much custom code for a `mdx-components.tsx` file, consider moving to a dedicated auto-type-table.mdx file or similar

yamcodes:
Addressed the review comments:

1. **Too much code in `mdx-components.tsx`** — Extracted the `AutoTypeTable` logic into a dedicated server component at `~/components/ui/auto-type-table.tsx`. The `mdx-components.tsx` now simply imports and re-exports it.

2. **`AutoTypeTable` crash at runtime** — The new component is a proper async React Server Component that awaits `generator.generateTypeTable({ name, path, type }, options)` (the correct v5.x API), runs markdown and type rendering via `markdownRenderer`, and passes the resolved React nodes into `TypeTable` with `expandAll`.

3. **`@radix-ui/react-collapsible` transitive dep** — Added it as a direct dependency in `apps/www/package.json` (along with `hast-util-to-jsx-runtime`, `remark`, and `remark-rehype` that the new component uses).

Commits: a27993d5

PR #1262, opened 2026-06-29
feat(nuxt): enforce strict-layout Vite boundary for userland server imports

### LINKED ISSUE #1244: feat(nuxt): enforce strict-layout Vite boundary for userland server imports
## Summary

This is the final, isolated security patch for the Nuxt parity epic. ADR-0013 requires the strict layout to block client-side imports of **any** server-only schema file, including userland files like `~/env/server.ts`. The current Vite plugin only blocks the `@arkenv/nuxt/server` package entry point, so the boundary is incomplete.

## Current behavior

The Vite plugin registered by `@arkenv/nuxt/module` blocks:

- `id === "@arkenv/nuxt/server"`
- any file under `@arkenv/nuxt/src|dist` named `server`

It does **not** block imports of the user's own `env/server.ts` (or any file ending with `/server` inside the configured schema directory), which means a developer can still import server-only schemas into a client component and leak server variable names/types to the browser bundle.

## Desired behavior

Extend the Vite `resolveId` plugin so that, during the client build, it also rejects module IDs whose resolved path ends with `/server` inside the schema directory. The error message should remain the existing branded message:

```
[ArkEnv] Importing server-only environment schema on the client is not allowed!
```

The check should work for both absolute paths and Nuxt aliases that resolve into the schema directory (e.g., `~/env/server`, `~~/env/server`).

## Acceptance criteria

- [ ] The Vite plugin blocks imports of userland `env/server.ts` (or any file ending with `/server` inside the configured schema directory) during the client build.
- [ ] The plugin still blocks `@arkenv/nuxt/server` and package-internal server entry points.
- [ ] The error message is the existing branded ArkEnv message.
- [ ] Server-side imports of the same file are unaffected.
- [ ] A test verifies that a client-side import of `~/env/server.ts` fails the build, while a server-side import succeeds.

## Out of scope

- Flat layout parsing and runtime API changes.
- Build-time validation and codegen.
- Changing the error message text or adding a remediation hint (can be a follow-up).

## References

- Tracking epic: #1241
- ADR-0013 — Nuxt strict-layout compile-time boundary via Vite plugin
- `@arkenv/nuxt/src/module.ts` — current Vite plugin implementation

yamcodes:
Fixes #1244

Extend the Vite plugin registered by "@arkenv/nuxt/module" so it also rejects client-side imports of userland server-only schema files inside the configured strict-layout schema directory. The existing branded error message is preserved, and server-side builds remain unaffected.

### DIFF
--- a/.changeset/nuxt-strict-userland-boundary.md
+++ b/.changeset/nuxt-strict-userland-boundary.md
@@ -1,0 +1,13 @@
+---
+"@arkenv/nuxt": patch
+---
+
+#### Enforce strict-layout Vite boundary for userland server imports
+
+Extend the `@arkenv/nuxt/module` Vite plugin to block client-side imports of userland server-only schema files inside the configured strict-layout schema directory. The plugin now rejects any resolved module path ending with `/server` under the schema base directory, including imports via Nuxt aliases such as `~/env/server` and `~~/env/server`.
+
+Server-side builds remain unaffected, and the existing branded error message is preserved:
+
+```
+[ArkEnv] Importing server-only environment schema on the client is not allowed!
+```
--- a/packages/nuxt/README.md
+++ b/packages/nuxt/README.md
@@ -77,4 +77,4 @@ This allows you to safely swap public configuration values in production without
 
 ## Client-Side Security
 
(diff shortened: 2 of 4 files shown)


yamcodes:
LGTM

PR #1278, opened 2026-07-07
docs(arkenv): fix options type resolution by exporting RuntimeEnvironment

yamcodes:
### Summary
This PR fixes an issue where the type of the `env` property in `ArkEnvConfig` was documented as `any` in the generated type table within `options.mdx`. 

### Cause
`env`'s type is `RuntimeEnvironment`, which was an unexported type alias resolving to `Dict<string>`. `Dict` was imported from the internal monorepo package `@repo/types`. When documentation generators (like `fumadocs-typescript`) compile the code in isolation, any failure to resolve the `@repo/types` package results in the type defaulting to `any`.

### Solution
1. Defined `RuntimeEnvironment` directly as `Record<string, string | undefined>` instead of importing `Dict` from `@repo/types`.
2. Formally exported `RuntimeEnvironment` so that it is part of `arkenv`'s public API surface and can be resolved correctly by documentation generators.
3. Created a changeset for the patch version bump.


yamcodes:
Not sure we need this, checking

Issue #1266, opened 2026-07-05
Support hosting presets in arkenv init (Phase 1)


yamcodes:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Support hosting presets in `arkenv init` (Phase 1)

**Current behavior:**
`arkenv init` creates a boilerplate configuration (`env.ts`) with default variables (`DATABASE_URL`, `NEXT_PUBLIC_API_URL`, etc.). It does not offer any hosting-specific environment variables.

**Desired behavior:**
1. During `arkenv init`, prompt the user: "Select a hosting provider preset (optional)" with options:
   - `None` (default)
   - `Vercel`
   - `Netlify`
2. If a provider is selected:
   - Add their standard environment variables to the generated schema in `env.ts` (using correct framework prefixes e.g. `NEXT_PUBLIC_` for Next.js, `NUXT_PUBLIC_` for Nuxt, `VITE_` for Vite).
   
**Preset Schemas (for ArkType validator):**
- **Vercel:**
  - `VERCEL: "string?"`
  - `VERCEL_ENV: "'production' | 'preview' | 'development'?"`
  - `VERCEL_URL: "string?"`
  - `<prefix>VERCEL_ENV: "'production' | 'preview' | 'development'?"`
  - `<prefix>VERCEL_URL: "string?"`
- **Netlify:**
  - `NETLIFY: "string?"`
  - `CONTEXT: "'production' | 'deploy-preview' | 'branch-deploy'?"`
  - `URL: "string?"`
  - `DEPLOY_URL: "string?"`
  - `<prefix>CONTEXT: "'production' | 'deploy-preview' | 'branch-deploy'?"`
  - `<prefix>URL: "string?"`

*(Note: Generate equivalent schemas for Zod and Valibot validators based on the target validator chosen during init).*

**Key interfaces:**
- `ProjectOptions` type (add optional `hostPreset` field).
- Prompts in `init` command (use `@clack/prompts` select).
- Templates under `packages/arkenv/src/features/scaffold/templates/` (add hostPreset variables to templates).

**Acceptance criteria:**
- [ ] Running `arkenv init` prompts the user for a hosting provider.
- [ ] Selecting `Vercel` or `Netlify` adds the corresponding preset variables to the generated schema.
- [ ] Generated client-side variables dynamically use the detected framework's prefix.
- [ ] Core package runtime remains completely untouched.

abhimanyutiwaribot:
[...]

Issue #610, opened 2025-12-23
Presets


yamcodes:
Support presets for popular hosting services, inspired by [Envin](https://envin.turbostarter.dev/docs/customization#extending-presets).

Depending on how much this can add to the core library, consider adding these in a separate library. Not everybody needs those.

aruaycodes:
> @aruaycodes would you like to take this?

I will take it

aruaycodes:
I think I am going to work on the CLI-tool, as that one sounds more fit to what "ArkEnv" needs

yamcodes:
Great work @abhimanyutiwaribot . Closed

Issue #181, opened 2025-09-19
Support Bun frontend `process.env` prefixing (e.g. `BUN_PUBLIC_*`)


yamcodes:
When using arkenv in a Bun + React project, process.env is not defined in the frontend runtime. Bun only exposes environment variables that are explicitly prefixed (default: BUN_PUBLIC_*) and accessing process.env directly throws a runtime error.

Additionally, during the build step, Bun's bundler defaults to browser builds, which causes import errors such as:

```ts
import { styleText } from "util";
             ^
error: Browser polyfill for module "node:util" doesn't have a matching export named "styleText"
```

### Steps to Reproduce:
1. Create a clean React template with bun init.
2. Install arkenv.
3. Import and call createEnv.
4. Run bun serve → runtime error: process is not defined.
5. Run bun build → bundler error on styleText.

### Expected behavior:
arkenv should support Bun's process.env model, allowing retrieval of all variables with a given prefix (e.g., `BUN_PUBLIC_*`).

### Workaround:
Manually construct the env object and pass only the prefixed vars to the schema:

```ts
import { type } from "arktype"

const envType = type({ MY_VAR: "string" })
const env = envType.assert({ MY_VAR: process.env.BUN_PUBLIC_MY_VAR })
export default env
```

### Related:
* Bun issue: https://github.com/oven-sh/bun/issues/19313
* #288 (Vite frontend `import.meta.env` prefixing)

### Action Items:
* Add support for collecting env vars by a configurable prefix (default: `BUN_PUBLIC_*`).
* Ensure builds targeting browser/bun don't attempt to import Node-only utilities like `util.styleText`.

yamcodes:
Note that this could also be relevant for Vite and Next.js

PR #1288, opened 2026-07-10
Add hosting presets to init CLI

abhimanyutiwaribot:
Supporting hosting presets in **arkenv init** (Phase 1)
This PR implements Phase 1 of the hosting presets (#610 / #1266) to offer opt in Vercel/Netlify variables during intialization.

#### What Changed 
1. CLI prompt: Added hosting preset step to the arkenv init wizard (None / Vercel / Netlify).
2. Static Codegen: Created static type mapping logic for Vercel/Netlify preset keys, automatically adapts validation enums to Zod, Valibot, and ArkType schemas.
3. Framework Prefix Adaptation: Prefixing applies client variables dynamically based on the active framework (e.g. NEXT_PUBLIC_, VITE_, NUXT_PUBLIC_).
4. Planner integration: Combines preset keys into .env and .env.example boilerplate files.
### Verification
Added prompt, template, and planner unit tests. All 278 test cases pass successfully.
Screenshot 2026-07-10 232054

### DIFF
--- a/.changeset/add-hosting-presets-init.md
+++ b/.changeset/add-hosting-presets-init.md
@@ -10,7 +10,7 @@ Usage:
 
 1. Run initialization in an existing project:
    ```bash
-   npx arkenv init
+   npx @arkenv/cli@latest init
    ```
 2. Select Vercel or Netlify from the "Select a hosting provider preset (optional)" prompt.
 3. The generated environment configuration file (`env.ts`) will contain pre-typed environment variables, automatically prefixing them based on the active framework (e.g. `NEXT_PUBLIC_VERCEL_ENV` and `NEXT_PUBLIC_VERCEL_URL` for Next.js).
--- a/.changeset/add-hosting-presets-init.md
+++ b/.changeset/add-hosting-presets-init.md
@@ -1,0 +1,16 @@
+---
+"@arkenv/cli": patch
+---
+
+#### Add hosting presets option to init CLI
+
+Support Vercel and Netlify presets when initializing a project via `arkenv init`.
+
+Usage:
+
+1. Run initialization in an existing project:
+   ```bash
+   npx arkenv init
+   ```
+2. Select Vercel or Netlify from the "Select a hosting provider preset (optional)" prompt.
+3. The generated environment configuration file (`env.ts`) will contain pre-typed environment variables, automatically prefixing them based on the active framework (e.g. `NEXT_PUBLIC_VERCEL_ENV` and `NEXT_PUBLIC_VERCEL_URL` for Next.js).
--- a/packages/cli/src/cli/ui/prompts/wizard.ts
+++ b/packages/cli/src/cli/ui/prompts/wizard.ts
@@ -13,7 +13,7 @@ type HasTypeFileAtPath = (options: {
 }) => boolean | Promise<boolean>;
 
 type ExistingProjectDefaults = Partial<
-	Pick<ProjectOptions, "framework" | "bunFeatures" | "disableCodegen">
+	Pick<ProjectOptions, "framework" | "bunFeatures" | "disableCodegen" | "hostPreset">
 > & {
 	defaultEnvPath?: string;
 	tsConfig?: ParsedTsConfig | null;
@@ -44,6 +44,7 @@ export async function runPromptWizard(
 			| "framework"
 			| "bunFeatures"
 			| "disableCodegen"
+			| "hostPreset"
 		>
 	> & {
 		examples?: Example[];
@@ -212,6 +213,7 @@ async function runExistingProjectWizard(
 			envKeys: detectedKeys ?? undefined,
 			disableCodegen: defaults?.disableCodegen ?? false,
 			wrapNextjsConfig: framework === "nextjs" ? true : undefined,
+			hostPreset: defaults?.hostPreset || "none",
 		});
 	}
 
@@ -322,6 +324,13 @@ async function runExistingProjectWizard(
 		// 8. validator
 		const validator = unwrapPrompt(await steps.validator());
 
+		// 8b. hostPreset
+		const hostPreset = unwrapPrompt(
+			await steps.hostPreset({
+				initialValue: defaults?.hostPreset,
+			}),
+		);
+
 		// 9. useEnvExample
 		const useEnvExample = unwrapPrompt(
 			await steps.useEnvExample({
@@ -346,6 +355,7 @@ async function runExistingProjectWizard(
 			installTypeDefinitions,
 			envDtsHandling,
 			validator,
+			hostPreset,
 			bunFeatures,
 			language: "ts",
 			installSkill: false,
--- a/packages/cli/src/cli/ui/prompts/wizard.ts
+++ b/packages/cli/src/cli/ui/prompts/wizard.ts
@@ -324,14 +324,16 @@ async function runExistingProjectWizard(
 		// 8. validator
 		const validator = unwrapPrompt(await steps.validator());
 
-		// 8b. hostPreset
-		const hostPreset = unwrapPrompt(
-			await steps.hostPreset({
-				initialValue: defaults?.hostPreset,
-			}),
-		);
+		// 9. hostPreset
+		const hostPreset = defaults?.hostPreset !== undefined
+			? defaults.hostPreset
+			: unwrapPrompt(
+				await steps.hostPreset({
+					initialValue: defaults?.hostPreset,
+				}),
+			);
 
-		// 9. useEnvExample
+		// 10. useEnvExample
 		const useEnvExample = unwrapPrompt(
 			await steps.useEnvExample({
 				detectedKeys,
--- a/packages/cli/src/features/scaffold/env-template.ts
+++ b/packages/cli/src/features/scaffold/env-template.ts
@@ -1,5 +1,6 @@
 import type { ProjectOptions } from "./plan";
 import { arktypeTemplate, valibotTemplate, zodTemplate } from "./templates";
+import { getPresetKeys, getFieldDefinition } from "./templates/presets";
 
 /**
  * Generate the complete environment configuration template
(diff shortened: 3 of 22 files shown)


.changeset/add-hosting-presets-init.md
```diff
@@ -0,0 +1,16 @@
+---
+"@arkenv/cli": patch
```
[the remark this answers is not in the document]
yamcodes: No it doesn't, we're in v0 for the CLI

packages/cli/src/features/scaffold/templates/presets.ts
yamcodes, line 1: This is not really scalable - consider getting the templates from a standard TS config object (with a clear interface) or a json.

As it is, in order to add a new preset (say, Cloudflare) we need to access `getPresetKeys` and `getFieldDefinition` separately (instead of making an addition in one place).

.changeset/add-hosting-presets-init.md
```diff
@@ -0,0 +1,16 @@
...
+#### Add hosting presets option to init CLI
+
+Support Vercel and Netlify presets when initializing a project via `arkenv init`.
+
+Usage:
+
+1. Run initialization in an existing project:
+   ```bash
+   npx arkenv init
+   ```
```
yamcodes, line 14: Correction, it's `npx @arkenv/cli@latest init`

packages/cli/src/features/scaffold/templates/presets.ts
yamcodes, line 1: This is not really scalable, right now in order to have a new preset we have to add it in both `getPresetKeys` and `getFieldDefinition` and access code that feels very static (I would expect adding new presets to be a little more dynamic).

Instead, consider a more schema-like solution, like reading these presets as json or at least a TS config object with a clear interface.

packages/cli/src/features/scaffold/plan.ts
```diff
@@ -37,6 +37,7 @@ export type ProjectOptions = {
 	envExampleContent?: string;
 	envContent?: string;
 	gitignoreContent?: string;
+	hostPreset?: "none" | "vercel" | "netlify";
```
yamcodes: If we add a new preset we have to remember going to `plan.ts` to `ProjectOptions` and add it here.

Instead I'd expect this type to come from a shared place (along with the presets) so the change of adding a new preset happens in one place.

packages/cli/src/features/scaffold/planner.ts
```diff
@@ -400,13 +401,20 @@ export function createPlan(state: CollectedState): ScaffoldingPlan {
 	const hasEnv = existingFiles.includes(envPath);
 	const hasEnvExample = existingFiles.includes(envExamplePath);
 
+	const clientPrefix =
+		options.framework === "nuxt" ? "NUXT_PUBLIC_" : (options.framework === "nextjs" ? "NEXT_PUBLIC_" : getFrameworkPrefix(options.framework));
+	const presetKeys = options.hostPreset ? getPresetKeys(options.hostPreset, clientPrefix) : [];
+	const combinedEnvKeys = options.envKeys || presetKeys.length > 0
+		? Array.from(new Set([...(options.envKeys || []), ...presetKeys]))
+		: undefined;
```
yamcodes: This code is a bit too complicated and repetitive for what it needs to do - the `getFrameworkPrefix` function by itself should already provide the prefix, we don't need these custom ternaries here

packages/cli/src/features/scaffold/env-template.ts
```diff
@@ -75,16 +76,19 @@ export function getStrictEnvTemplates(
 	const pkgName =
 		options.framework === "nuxt" ? "@arkenv/nuxt" : "@arkenv/nextjs";
 
+	const existingKeys = envKeys && envKeys.length > 0
+		? envKeys
+		: (options.framework === "nuxt" ? ["DATABASE_URL", "NUXT_PUBLIC_API_URL", "NODE_ENV"] : ["DATABASE_URL", "NEXT_PUBLIC_API_URL", "NODE_ENV"]);
```
yamcodes: This line doesn't feel great, it hardcodes a list of envs per frameworks. This should come from a single place (like `exampleEnvDefaults`)

packages/cli/src/cli/ui/prompts/wizard.ts
```diff
@@ -322,6 +324,13 @@ async function runExistingProjectWizard(
 		// 8. validator
 		const validator = unwrapPrompt(await steps.validator());
 
+		// 8b. hostPreset
```
yamcodes: (nitpick) why not 9? (we can move the next ones to be 10 and up)

yamcodes (commented):
This is a welcome contribution! Thank you!

There are just a few areas that violate the "Locality of Change" principle. Although, the current code in `dev` already largely violates anyway so I'd understand if you push back here. I'm going to file a separate issue to address this in `v1` so we can do what we can here (to at least make the greenfield code respect locality of change) and fix the existing patterns separately.

Edit: see https://github.com/yamcodes/arkenv/issues/1289

packages/cli/src/features/scaffold/templates/presets.ts
abhimanyutiwaribot: yep so, a single declarative configuration object (PRESETS),  for future presets.

yamcodes: This is a duplicate of https://github.com/yamcodes/arkenv/pull/1288#discussion_r3561291055, resolving

packages/cli/src/features/scaffold/templates/presets.ts
```diff
@@ -1,6 +1,61 @@
-import type { Framework, Validator } from "../plan";
+type Validator = "arktype" | "zod" | "valibot";
+type Framework = "vite" | "bun-fullstack" | "vanilla" | "nextjs" | "nuxt";
 
-export type HostPreset = "none" | "vercel" | "netlify";
+export const PRESETS = {
```
yamcodes, line 15: a great change in the right direction - can we make this satisfy a typed interface so it's easier to maintain? (as opposed to a free-form const)

packages/cli/src/features/scaffold/templates/presets.ts
```diff
@@ -1,6 +1,61 @@
-import type { Framework, Validator } from "../plan";
+type Validator = "arktype" | "zod" | "valibot";
+type Framework = "vite" | "bun-fullstack" | "vanilla" | "nextjs" | "nuxt";
```
yamcodes: this is fine but technically that info already exists elsewhere, consider inheriting this list from that. not a big deal since new frameworks rarely get added

packages/cli/src/features/scaffold/templates/presets.ts
```diff
@@ -1,6 +1,61 @@
-import type { Framework, Validator } from "../plan";
+type Validator = "arktype" | "zod" | "valibot";
```
yamcodes: this is fine but technically this info already exists elsewhere as a SoT, considering inheriting this list from that. not a big deal since new validators rarely (if ever) will get added to this list.

packages/cli/src/features/scaffold/templates/presets.ts
```diff
@@ -1,6 +1,61 @@
...
+			},
+			VERCEL_ENV: {
+				arktype: `"'production' | 'preview' | 'development'?"`,
+				zod: `z.enum(["production", "preview", "development"]).optional()`,
+				valibot: `v.optional(v.picklist(["production", "preview", "development"]))`,
+			},
+			VERCEL_URL: {
+				arktype: `"string?"`,
+				zod: "z.string().optional()",
+				valibot: "v.optional(v.string())",
```
yamcodes: I'm not really sure how I feel about holding validator definitions as strings here.

Can we think of an alternative? Like storing them as true Valibot/Zod/Arktype definitions and inheriting/injecting them where needed? Or maybe a more complex object that can contain them as not-strings.

See how they did it in T3 env:

https://github.com/t3-oss/t3-env/blob/main/packages/core/src/presets-zod.ts

packages/cli/src/features/scaffold/templates/presets.ts
```diff
@@ -0,0 +1,81 @@
...
+export type HostPreset = "none" | "vercel" | "netlify";
+
+/**
+ * Get the framework-specific client prefix.
+ */
+export function getFrameworkPrefix(framework?: Framework): string {
+	if (framework === "nextjs") return "NEXT_PUBLIC_";
+	if (framework === "nuxt") return "NUXT_PUBLIC_";
+	if (framework === "vite") return "VITE_";
+	if (framework === "bun-fullstack") return "BUN_PUBLIC_";
```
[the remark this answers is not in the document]
yamcodes: it's the default, see: https://github.com/yamcodes/arkenv/issues/181

packages/cli/src/features/scaffold/templates/presets.ts
abhimanyutiwaribot: i think in t3-env they store presets as runtime schema functions.
In our case, storing actual runtime Zod/Valibot/ArkType schema objects in the CLI would bloat the CLI package with unnecessary dependencies and require complex AST serialization to write the generated code to `env.ts`.

i think we can implement a structured, meta-data schema using typed definitions.

```
fields: {
  VERCEL: { type: "string" },
  VERCEL_ENV: { type: "enum", values: ["production", "preview", "development"] }
}
```
so during code generation, `getFieldDefinition` func dynamically compiles these typed descriptoes into validators code strings, and enforces type safety through Typed interfaces.

yamcodes: ok , after thinking about it a bit, i like your suggestion as a pragmatic tradeoff for `dev`. i think it accomplishes all of our goals without introducing too much bloat.

i noticed that it kinda rhymes with a JSON Schema subset. meaning, if we used json schema we could accomplish this preset approach without a bespoke schema. since we have two kinds only (`string` and `enum`) it's fine for now, but we should keep an eye on it if it grows - consider switching to json schema or something like that if needed then.

constraint i'd hold: don't grow `PresetField` into an ad-hoc schema language. if presets need more than string/enum, we should either (a) commit to a real JSON Schema subset as the IR, or (b) on `v1` keep PRESETS as host semantics only and let dialects own rendering - not invent a third vocab.

i'll add it as an ADR.

yamcodes:
@all-contributors please add @abhimanyutiwaribot for code, doc, example, test, tool

PR #1311, opened 2026-07-12
fix: (v1) Refactor CLI scaffolding using Strategy pattern for Validators and Frameworks

### LINKED ISSUE #1289: (v1) Refactor CLI scaffolding using Strategy pattern for Validators and Frameworks

## Problem Statement
In the CLI scaffolding feature, `planner.ts` and `env-template.ts` have grown too large (way over the comfortable limit of 200 LOC) and contain nested `if` and `switch` statements with repeated logic to handle various validator and framework combinations. 
This code structure violates the locality of change principle. Adding support for a new framework or validator integration requires modifying multiple distant, unexpected internal places (e.g. `planner.ts`, `env-template.ts`, templates, etc.), making it hard to maintain and extend.

## Desired Behavior
Refactor the scaffolding planning logic by extracting validator-specific and framework-specific operations into individual strategy files under `packages/cli/src/features/scaffold/validators/` and `packages/cli/src/features/scaffold/frameworks/`.

To resolve the M x N Matrix Problem (where frameworks and validators overlap, such as client prefixes and file layout structures):
1. **Context Sharing**: Introduce a shared `ScaffoldContext` containing framework configurations (e.g., `clientPrefix`) and options. Pass this context into the Validator Strategy methods.
2. **Inversion of Control for Templates**: Let the `FrameworkStrategy` coordinate the file layout and request the appropriate templates from the `ValidatorStrategy` (e.g., `getSimpleTemplate` or `getStrictTemplates`), rather than letting the validator strategy determine the file layout.
3. **Static Registries**: Register strategies in static maps (`VALIDATORS` and `FRAMEWORKS`) enforcing exhaustiveness checks via TypeScript.

Both `planner.ts` and `env-template.ts` should be reduced in size (targeting < 200 lines each) by delegating to these strategy modules.

## Acceptance Criteria
- [ ] Create `Strategy` interfaces or types for both Validators and Frameworks.
- [ ] Create individual files for validators: `arktype`, `zod`, and `valibot` under `packages/cli/src/features/scaffold/validators/`.
- [ ] Create individual files for frameworks: `vite`, `bun-fullstack`, `nextjs`, `nuxt`, and `vanilla` under `packages/cli/src/features/scaffold/frameworks/`.
- [ ] Register all validator and framework strategies in static maps (registries) with exhaustive type enforcement.
- [ ] Update `planner.ts` and `env-template.ts` to delegate to these strategies, reducing their LOC to < 200 lines.
- [ ] Ensure all existing unit tests in `packages/cli/src/features/scaffold/` pass without modification (or with minor test setup changes if internal APIs changed, but no functional/behavioral regressions).
- [ ] Verify that running scaffolding produces identical results/configs/dependencies.

yamcodes:
Fixes #1289

Refactor CLI scaffolding planning and template generation to use validator and framework strategy registries, implemented **natively on `v1`** under `packages/arkenv/src/features/scaffold/`.

Previously this PR targeted `dev` at `packages/cli/` — that was incorrect for this `(v1)`-labeled issue. This branch has been rebased onto `v1` with v1-specific adaptations:
- Strict-layout shared schema uses `@arkenv/core` (not `@arkenv/nextjs/shared`)
- Multiline quoted value stripping in `stripValuesFromEnvContent`
- Independent `.env` / `.env.local` gitignore matching

`planner.ts` and `env-template.ts` now delegate to strategy registries (20 and 47 LOC respectively).

## Test plan
- [x] `pnpm --filter arkenv run typecheck`
- [x] `pnpm --filter arkenv exec vitest run src/features/scaffold` (105 tests)
- [x] `pnpm run fix`


.changeset/scaffold-strategy-refactor.md
yamcodes, line 1: if it's internal only it doesn't need a changeset as per `/changeset` skill

packages/arkenv/src/features/scaffold/frameworks/shared.ts
```diff
@@ -0,0 +1,228 @@
...
+		if (key === "NODE_ENV") {
+			defaults[key] = "development";
+		} else if (key === "PORT") {
+			defaults[key] = "3000";
+		} else if (key === "DATABASE_URL") {
+			defaults[key] = "postgres://localhost:5432/mydb";
+		} else {
+			defaults[key] = "";
+		}
+	}
```
yamcodes, line 41: Any way to avoid an if-else chain? Maybe a switch/case block or ts-pattern?

packages/arkenv/src/features/scaffold/frameworks/shared.ts
yamcodes, line 1: this file is good but there are many repeated tokens i'd consider consolidating or saving as consts (as opposed to "magic strings") like `@/*`, etc.

yamcodes (commented):
some of the files in this pr have imports containing "../", that is usually a sign to use absolute imports like "@/" or "~/".

also some stuff is missing jsdoc - run the `/jsdoc` skill.

i'd also run `/typescirpt` on this once, and maybe even `/improve-codebase-architecture`

.changeset/scaffold-strategy-refactor.md
yamcodes: Agreed — removed the changeset. This is an internal-only refactor with no consumer-facing change.

packages/arkenv/src/features/scaffold/frameworks/shared.ts
yamcodes: Addressed — replaced the if/else chain with a lookup map (`ENV_KEY_DEFAULTS`). Skipped `ts-pattern` since it is not a dep here and a map is enough for this case.

packages/arkenv/src/features/scaffold/frameworks/shared.ts
yamcodes: Addressed — extracted repeated tokens into named constants (`TS_CONFIG_ALIAS_KEY`, `GENERATED_ENV_DIR`, `GENERATED_ENV_MODULE`, strict-layout basenames/labels, etc.) and deduplicated the codegen import-path guard into `resolveCodegenImportPath`.

PR #1325, opened 2026-07-16
feat(cli): machine-actionable error codes in `--agent` JSON output

### LINKED ISSUE #1313: CLI: machine-actionable error codes in `--agent` JSON output

## Summary

When `arkenv init` refuses to proceed because a safety check tripped (unmet technical requirements, dirty git tree, non-empty directory), the `--agent` / `--json` output does not tell the calling agent *which* check failed or how to escalate. Agents are left pattern-matching on prose like "Use --force to bypass these checks."

Spun out of the discussion on #1111, where it was decided that `--agent` must never imply `--force` — meaning agents need a deliberate, machine-readable path to escalate instead.

## Current behavior

- Safety-check refusals are reported via human-oriented `error`/`info` log lines (e.g. `Technical requirements not met:` … `Use --force to bypass these checks.`), and the command exits without a structured explanation.
- The JSON reporter's terminal `status: "error"` payload carries only free-text `message`/`error` fields — no stable identifier distinguishing "requirements not met" from "dirty git tree" from any other failure.

## Desired behavior

Every deliberate refusal in the CLI's JSON output carries:

- a stable, documented error `code` (e.g. `REQUIREMENTS_NOT_MET`, `GIT_TREE_DIRTY`, `NON_EMPTY_DIR`) that is part of the CLI's public contract;
- a `retryWith` hint listing the flags that would bypass the check (e.g. `["--force"]`), empty when the failure is not bypassable;
- enough structured detail for the agent to report the problem (e.g. the failed requirements with `current`/`expected`).

Unexpected/internal errors keep a generic code (e.g. `INTERNAL`) so consumers can distinguish "the CLI refused" from "the CLI broke".

## Acceptance criteria

- [ ] Each safety-check refusal in `init` (requirements, git status, non-empty directory) emits a distinct stable code in JSON mode
- [ ] Bypassable refusals include a `retryWith` array naming the bypass flag(s)
- [ ] Codes are documented (CLI docs and/or agent skill guidance) as a stable contract
- [ ] The agent skill guidance explains the escalation pattern: run without `--force`, inspect `code`/`retryWith`, retry deliberately
- [ ] Human-readable (non-JSON) output is unchanged
- [ ] Tests cover the JSON shape for each refusal path

## Out of scope

- Changing `--agent`, `--yes`, or `--force` semantics (settled in #1110 / #1111)
- Error codes for validation failures (`ArkEnvError`) — separate surface
- Retrofitting codes onto commands other than `init` (can follow once the pattern exists)

## Related

- #1111 — decision: `--agent` implies `--yes`, never `--force`
- #1110 — decision: keep both `--yes` and `--force`

yamcodes:
Fixes #1313

## Summary

Adds a stable, machine-actionable error contract to the CLI's `--json` / `--agent` output so calling agents can tell *which* safety check refused and *how* to escalate, instead of pattern-matching on human-oriented prose.

Every deliberate `init` refusal now emits a structured payload on `stdout` with a stable `code` and a `retryWith` hint. Human-readable (non-JSON) output is byte-for-byte unchanged (the existing `error`/`info` lines still go to `stderr` in JSON mode).

## Error codes (public contract)

| `code` | When | `retryWith` |
| --- | --- | --- |
| `REQUIREMENTS_NOT_MET` | A technical requirement failed (includes per-requirement `details` with `current`/`expected`) | `["--force"]` |
| `GIT_TREE_DIRTY` | The git working tree is not clean | `["--force"]` |
| `NON_EMPTY_DIR` | The target directory is not empty | `["--force"]` |
| `INTERNAL` | An unexpected failure — the CLI *broke* rather than *refused* | `[]` |

A non-empty `retryWith` marks a bypassable refusal; an empty one (and `INTERNAL`) means retrying with flags won't help.

## Changes

- New `packages/cli/src/shared/errors.ts` with `ERROR_CODES` + `Refusal` type (the public contract).
- New `refuse(refusal)` method on the reporter port / `LoggerPort`, implemented across the JSON (structured stdout), text/silent (no-op), and memory (test) reporters.
- `init` safety-check paths now emit the matching code alongside the unchanged human output.
- `fatal` now carries the generic `INTERNAL` code so consumers can distinguish "refused" from "broke".
- Docs: CLI docs (`Machine-readable output` section) and the ArkEnv agent skill document the codes and the escalation pattern (run without `--force`, inspect `code`/`retryWith`, retry deliberately).
- Tests cover the exact JSON shape for each refusal path; verified end-to-end against the built CLI.

## Acceptance criteria

- [x] Each safety-check refusal emits a distinct stable code in JSON mode
- [x] Bypassable refusals include a `retryWith` array naming the bypass flag(s)
- [x] Codes documented as a stable contract (CLI docs + agent skill)
- [x] Agent skill explains the escalation pattern
- [x] Human-readable output unchanged
- [x] Tests cover the JSON shape for each refusal path

### DIFF
--- a/.changeset/machine-actionable-error-codes.md
+++ b/.changeset/machine-actionable-error-codes.md
@@ -1,21 +1,10 @@
 ---
-"@arkenv/cli": minor
+"@arkenv/cli": patch
 ---
 
-#### Add machine-actionable error codes to `init` JSON output
+#### Emit machine-actionable error codes in `init` JSON output
 
-**BREAKING CHANGE**: The `--json` / `--agent` terminal error payload now always carries top-level `code`, `message`, and `retryWith` fields. For unexpected failures, `message` moved out of `details` to the top level: the shape changed from `{ status: "error", details: { message, error } }` to `{ status: "error", code: "INTERNAL", message, retryWith: [], details: { error } }`. Consumers that read `details.message` must now read the top-level `message`.
-
-In `--json` / `--agent` mode, every deliberate safety-check refusal now emits a stable, documented error `code` alongside a `retryWith` hint, so agents no longer have to pattern-match on prose to decide how to escalate. Human-readable (non-JSON) output is unchanged.
-
-- **`REQUIREMENTS_NOT_MET`** — a technical requirement failed. Includes per-requirement `details` with `current`/`expected`. `retryWith: ["--force"]`.
-- **`GIT_TREE_DIRTY`** — the git working tree is not clean. `retryWith: ["--force"]`.
-- **`NON_EMPTY_DIR`** — the target directory is not empty. `retryWith: ["--force"]`.
-- **`INTERNAL`** — an unexpected failure (the CLI broke rather than refused). `retryWith: []`.
-
-An empty `retryWith` means the refusal cannot be bypassed; a non-empty `retryWith` names the flag(s) to re-run with. Escalation pattern: run without `--force`, inspect `code`/`retryWith`, then retry deliberately.
-
-Example refusal payload written to `stdout`:
+Running `arkenv init --json` (or `--agent`) now tells you _why_ the CLI stopped, in a shape tools and AI agents can act on instead of scraping prose. Whenever a safety check refuses to proceed, the JSON written to `stdout` carries a stable `code` and a `retryWith` hint listing the flag(s) that would bypass the check. Human-readable (non-JSON) output is unchanged.
 
 ```json
 {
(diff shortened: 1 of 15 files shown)


yamcodes:
Addressed the review in 6020b2a9:

- Bumped the changeset from `patch` to **`minor`** (v0 convention for breaking changes; `major` is reserved for the v0→v1 transition).
- Documented the `fatal`/`INTERNAL` schema change as a `**BREAKING CHANGE**` in the changeset: `message` moves from `details.message` to the top level, so the payload goes from `{ status, details: { message, error } }` to `{ status, code: "INTERNAL", message, retryWith: [], details: { error } }`.

Kept the consistent top-level `{ code, message, retryWith, details? }` shape across both `refuse()` and `fatal()` rather than reverting, since establishing that unified contract is the goal of this PR.

yamcodes:
Update on the bump-type discussion (d16ef87b): after a closer look I've reverted this to a **`patch`** rather than `minor`.

Reasoning: the error JSON shape was never documented, and before this PR the safety-check refusals emitted **no JSON at all** (just stderr text + exit 1) — the only prior JSON error output came from `fatal` on unexpected internal crashes, which isn't a shape consumers build stable integrations on. On a v0 (0.3.x) package with an undocumented, crash-only error shape, this reads as an additive, non-breaking change.

I kept the shape refinement (`message` at the top level) as a low-key note at the bottom of the changeset for the rare consumer who parsed the old error output, rather than a `**BREAKING CHANGE**` banner.

PR #1352, opened 2026-07-17
docs: use AutoTypeTable for config options and expand twoslash coverage

yamcodes:
## Summary

Follows up on an audit of the [Next.js configuration page](https://arkenv-7pjewqvtn-yamcodes.vercel.app/docs/nextjs/configuration) for two recurring opportunities across the docs: replacing hand-written option tables with `AutoTypeTable`, and expanding `twoslash` coverage on config examples.

### `AutoTypeTable`
- Replace the hand-written "Configuration options" table on **`nextjs/configuration.mdx`** with `<AutoTypeTable path="../../packages/nextjs/src/config.ts" name="ArkEnvConfigOptions" />`. The type is already fully JSDoc-annotated, so the table now stays in sync with the source.
- Do the same on **`nuxt/configuration.mdx`** using `ModuleOptions`. Since that type had no JSDoc, add descriptions and `@default` tags to `ModuleOptions` in `packages/nuxt/src/module.ts` so the generated table keeps its descriptions/defaults.

### twoslash
- Add a `@arkenv/nextjs/config` entry to the twoslash path map in `lib/twoslash-options.ts`. Previously **every** `next.config.ts` example importing `withArkEnv`/`setupArkEnv` resolved to `any` (the module-resolution error was silently filtered), so twoslash produced no useful hovers.
- Convert the now-resolvable `next.config.ts` / `src/env.ts` examples to `ts twoslash` across:
  - `nextjs/configuration.mdx` (register, custom paths, `setupArkEnv` escape hatch)
  - `nextjs/faq.mdx` (`{ codegen: false }` wrapper + manual `createEnv`)
  - `nextjs/layouts/flat.mdx` and `nextjs/layouts/strict.mdx` (config wrapper steps)

Nuxt `nuxt.config.ts` blocks were intentionally left plain — `defineNuxtConfig` is a Nuxt auto-import and would error under twoslash without extra setup.

## Verification
- Ran the shared `arktypeTwoslashOptions` twoslasher (`scripts/twoslash-mdx.ts`) over every edited file: all new blocks compile, resolve `NextConfig`/`withArkEnv`/`ArkEnvConfigOptions` with correct hovers, and add **zero** new errors. (The pre-existing intentional error demo in `strict.mdx` is unchanged.)
- `biome check` clean on changed source files.

## Changeset
- `@arkenv/nuxt` patch — the `ModuleOptions` JSDoc improves editor DX on the public type.

### DIFF
--- a/.changeset/nuxt-module-options-jsdoc.md
+++ b/.changeset/nuxt-module-options-jsdoc.md
@@ -1,0 +1,19 @@
+---
+"@arkenv/nuxt": patch
+---
+
+#### Document `ModuleOptions` with JSDoc for better editor DX
+
+Add descriptions and `@default` tags to the `ModuleOptions` type so hovering `schemaPath`, `layout`, and `validate` in `nuxt.config.ts` surfaces inline documentation.
+
+```ts title="nuxt.config.ts"
+export default defineNuxtConfig({
+  modules: ["@arkenv/nuxt/module"],
+  arkenv: {
+    // Hovering these keys now shows their description and default value
+    schemaPath: "src/env.ts",
+    layout: "flat",
+    validate: true,
+  },
+});
+```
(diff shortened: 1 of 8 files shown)


yamcodes:
Addressed the nitpick in e142f34c — the `{ codegen: false }` example in the `> [!TIP]` callout now uses `ts twoslash`, matching every other `next.config.ts` block on the page. (It's the only fenced code inside a blockquote in the docs, so the Vercel preview build is the authoritative check that twoslash renders correctly inside the callout.)

PR #1361, opened 2026-07-17
docs: Forward-port ModuleOptions JSDoc from dev to v1

yamcodes:
Forward-ports the `@arkenv/nuxt` `ModuleOptions` JSDoc from dev PR #1352 (the only published-package change in that PR — the rest was `apps/www` docs, which aren't forward-ported).

## Change

Adds JSDoc descriptions + `@default` tags to `ModuleOptions` in `packages/nuxt/src/module.ts` (same path/package name on v1), so hovering the `arkenv` config keys in `nuxt.config.ts` surfaces inline documentation.

### v1 adaptations
- Preserved the v1-only fields `logger` and `logLevel` (part of the injectable-logger API that exists only on v1) and gave them concise JSDoc so the whole type is documented.
- Changeset uses the v1 package name — `@arkenv/nuxt` (unchanged between dev and v1).

## Verification
- `biome check` clean on the changed file and changeset.
- JSDoc-only change (comments): introduces **zero** new type errors — a `tsc --noEmit` on `packages/nuxt` yields the exact same environmental error set as pristine `origin/v1` (unbuilt `@repo/*` / `@arkenv/build` workspace deps); nothing references the edited declaration.

Source: dev PR #1352.

### DIFF
--- a/.changeset/nuxt-module-options-jsdoc-v1.md
+++ b/.changeset/nuxt-module-options-jsdoc-v1.md
@@ -1,0 +1,19 @@
+---
+"@arkenv/nuxt": patch
+---
+
+#### Document `ModuleOptions` with JSDoc for better editor DX
+
+Add descriptions and `@default` tags to the `ModuleOptions` type so hovering `schemaPath`, `layout`, `validate`, `logger`, and `logLevel` in `nuxt.config.ts` surfaces inline documentation.
+
+```ts title="nuxt.config.ts"
+export default defineNuxtConfig({
+  modules: ["@arkenv/nuxt/module"],
+  arkenv: {
+    // Hovering these keys now shows their description and default value
+    schemaPath: "src/env.ts",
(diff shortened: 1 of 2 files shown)


yamcodes:
Both accuracy notes addressed in 1efb5b9c:

- Dropped the misleading `@default "src/env.ts"` on `schemaPath` — reworded to describe the `findSchemaPath` auto-discovery instead.
- Dropped `@default "flat"` on `layout` and clarified that the layout is auto-detected from the schema structure (`strict` when the split files exist, `flat` as the fallback), matching `resolveLayout`'s behavior.

Kept `@default true` on `validate`, which is accurate (module `defaults.validate = true` / `options?.validate ?? true`).

Note: the original dev PR (#1352) carries the same misleading `@default` tags, so dev and v1 now differ slightly here — happy to open a small follow-up on `dev` to align it if wanted.

PR #1363, opened 2026-07-17
docs: clarify auto-detected schemaPath/layout in Nuxt ModuleOptions

yamcodes:
Follow-up to #1352 aligning `dev` with a review note from the v1 forward-port (#1361).

## Problem

The `ModuleOptions` JSDoc added in #1352 uses `@default "src/env.ts"` and `@default "flat"`, which are misleading:

- `schemaPath` is **not** literally defaulted — the module calls `findSchemaPath` to auto-discover among `env.ts`, `src/env.ts`, and strict-layout directories.
- `layout` is **auto-detected** by `resolveLayout` from the schema structure (`"strict"` when the split files exist, `"flat"` as the fallback), not defaulted to a fixed value.

## Change

- Drop the misleading `@default` tags on `schemaPath` and `layout`; reword the prose to describe the auto-detection accurately.
- Keep `@default true` on `validate` (accurate: `defaults.validate = true` / `options?.validate ?? true`).

This brings `dev` in line with the corrected v1 docs and keeps future forward-ports clean.

`@arkenv/nuxt` patch changeset included.

### DIFF
--- a/.changeset/nuxt-module-options-jsdoc-accuracy.md
+++ b/.changeset/nuxt-module-options-jsdoc-accuracy.md
@@ -1,0 +1,7 @@
+---
+"@arkenv/nuxt": patch
+---
+
+#### Clarify auto-detected `schemaPath` and `layout` in `ModuleOptions` docs
+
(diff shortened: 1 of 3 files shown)


yamcodes:
Addressed in a2ecef91 — aligned the sibling `ArkEnvConfigOptions` in `packages/nuxt/src/config.ts` too: dropped the `@default "src/env.ts"` / `@default "flat"` tags and reworded to describe the same `findSchemaPath` / `resolveLayout` auto-detection, keeping its `@example`, the `@deprecated` note on `"simple"`, and `@default true` on `validate`. The `@arkenv/nuxt` doc surface is now uniform.

PR #1385, opened 2026-07-17
feat(nuxt): Forward-port `arkenv` config-key typing from #1371 to v1

yamcodes:
Forward-ports #1371 (merged to `dev`) to the `v1` branch.

## Change

Types the `arkenv` module options key in `nuxt.config.ts` so consumers get autocomplete, type-checking, and JSDoc hovers instead of the loose index-signature fallback.

- Alias `ModuleOptions` to the documented `ArkEnvConfigOptions` (single source of truth for field-level JSDoc / `@default` tags).
- Augment `@nuxt/schema`'s `NuxtConfig` and `NuxtOptions` with `arkenv?: ModuleOptions` (lives in `module.ts`, ships in `dist/module.d.ts`).
- Add type-level regression tests covering the `ModuleOptions` alias and the `arkenv` config key (known keys accepted, unknown keys rejected).

## v1 adaptations

- v1's `ArkEnvConfigOptions` already covers `logger` / `logLevel`, so aliasing `ModuleOptions` to it picks those up automatically — the previously hand-maintained `ModuleOptions` fields are dropped in favor of the alias.
- Regression-test imports keep the v1 `arkenv` / `arkenvStandard` API instead of dev's `createEnv`.

## Verification

- `pnpm --filter @arkenv/nuxt run typecheck` passes (after building workspace deps).
- `vitest run --typecheck` on `@arkenv/nuxt`: **36 tests passed, no type errors**.
- `biome check` clean on the changed files and changeset.

Source: dev PR #1371 (`7d541b4e`).

### DIFF
--- a/.changeset/nuxt-augment-config-key.md
+++ b/.changeset/nuxt-augment-config-key.md
@@ -1,0 +1,20 @@
+---
+"@arkenv/nuxt": patch
+---
+
+#### Type the `arkenv` key in `nuxt.config.ts` via `@nuxt/schema` augmentation
+
+Augment `@nuxt/schema`'s `NuxtConfig` and `NuxtOptions` so the `arkenv` module options key is fully typed. Consumers now get autocomplete, type-checking, and JSDoc hovers for `arkenv` options directly in `nuxt.config.ts`, instead of falling back to a loose index signature.
+
+`ModuleOptions` is now an alias of the documented `ArkEnvConfigOptions`, so option hovers surface the existing JSDoc / `@default` tags from a single source of truth.
+
+```ts
+export default defineNuxtConfig({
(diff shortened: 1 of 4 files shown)


yamcodes:
Addressed in 8b4a7b53.

The alias dropped the field-level JSDoc because `ArkEnvConfigOptions` was undocumented on `v1`. I moved the deleted docs into `packages/nuxt/src/config.ts` (`schemaPath`, `layout`, `validate`) and added docs for the v1-only `logger` / `logLevel` fields, so the alias target is now the documented single source of truth and hovers on the `arkenv` config key surface JSDoc again.

Verified: `biome check` clean, and `vitest run --typecheck` on `@arkenv/nuxt` passes (36 tests, no type errors).

PR #1420, opened 2026-07-20
fix: security proxy prefers raw runtimeConfig/process.env strings over coerced values

### LINKED ISSUE #1327: fix(nuxt): security proxy prefers raw runtimeConfig/process.env strings over coerced values
## Summary

The Nuxt security proxy prefers raw string sources over the coerced validation target on **both** the client and server branches, so coercion (ADR 0002) is silently bypassed: a key typed as `number` returns a raw `string` at runtime.

## Details

In `packages/nuxt/src/create-env.ts` (~lines 327–372), the proxy getter resolves schema keys in this order:

- **Client branch:** `runtimeEnv` → `useRuntimeConfig().public` (raw strings) → `__NUXT__.config.public` (raw strings) → coerced target.
- **Server branch:** `runtimeEnv` → `runtimeConfig[prop]` / `runtimeConfig.public[prop]` (raw strings) → **raw `process.env[prop]`** → coerced target.

Severity on the server: `prop in process.env` is near-always true for schema keys (they were just validated from it), so the coerced target is effectively **unreachable in the common path** unless `runtimeEnv` explicitly owns the key. Server-side coercion is broken by default, not just in edge cases.

## Fix direction

Coerce values **before** they enter the runtime payload rather than shipping a validator to the client. Implementation note: Nitro applies `NUXT_PUBLIC_*` runtime overrides as **strings at boot**, so the fix needs a boot-time coercion pass (e.g. a Nitro plugin), not just build-time injection.

## Acceptance criteria

- [ ] A key declared with a coercing schema (e.g. `"string.numeric"` with coerce) returns a `number` on both server and client, including when overridden via `NUXT_PUBLIC_*` at boot.
- [ ] No validator code ships to the client bundle.
- [ ] The existing test "dynamically resolve client keys from useRuntimeConfig" is updated to assert coerced values.
- [ ] Regression tests cover the server branch's `process.env` fallback path.

## Landing

Lands on `dev` first (bug fix affecting current users), then forward-port to `v1` per the dual-tracking flow.

Part of #1105 (ADR 0015, #1326) — the aligned model's honesty claim is only as good as its weakest integration.

yamcodes:
Fixes #1327

## Summary
- Prefer the coerced validation target in the Nuxt security proxy instead of re-reading raw `useRuntimeConfig()` / `process.env` / `__NUXT__.config.public` strings (which silently undid ADR 0002 coercion).
- Update dynamic `useRuntimeConfig` tests to assert coerced `number`/`boolean` values, and add regression coverage for the server `process.env` and client `__NUXT__` paths.
- Add a missing `@arkenv/nuxt` Vitest project config so the package tests run in CI.

## Test plan
- [x] `pnpm exec vitest run --project @arkenv/nuxt`
- [x] `pnpm run typecheck`
- [x] `pnpm run test`
- [x] `pnpm run fix`

### DIFF
--- a/.changeset/fix-nuxt-proxy-coercion.md
+++ b/.changeset/fix-nuxt-proxy-coercion.md
@@ -2,9 +2,9 @@
 "@arkenv/nuxt": patch
 ---
 
-#### Fix security proxy returning raw strings instead of coerced values
+#### Fix number and boolean env values returning as strings
 
-Prefer the coerced validation target for schema-key reads instead of re-reading raw `useRuntimeConfig()`, `process.env`, or `__NUXT__.config.public` strings. Those sources still feed validation at create time, but serving them again on get silently undid coercion (for example a `number` key returning a `string`).
+Keep coerced types when reading from `env`. A key declared as `"number"` or `"boolean"` now returns a number or boolean at runtime, not the raw string from Nuxt runtime config.
 
 ```ts
 import { createEnv } from "@arkenv/nuxt";
@@ -14,5 +14,6 @@ export const env = createEnv({
   PORT: "number",
 });
 
-// env.NUXT_PUBLIC_PORT and env.PORT are numbers, not "3000"
-```
+// Was "3000" (string) — now 3000 (number)
+env.NUXT_PUBLIC_PORT;
+env.PORT;
--- a/.changeset/fix-nuxt-proxy-coercion.md
+++ b/.changeset/fix-nuxt-proxy-coercion.md
@@ -16,4 +16,5 @@ export const env = createEnv({
 
 // Was "3000" (string) — now 3000 (number)
 env.NUXT_PUBLIC_PORT;
+env.NUXT_PUBLIC_PORT;
 env.PORT;
--- a/.changeset/fix-nuxt-proxy-coercion.md
+++ b/.changeset/fix-nuxt-proxy-coercion.md
@@ -17,3 +17,4 @@ export const env = createEnv({
 // Was "3000" (string) — now 3000 (number)
 env.NUXT_PUBLIC_PORT;
 env.PORT;
+```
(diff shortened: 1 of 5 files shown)


.changeset/fix-nuxt-proxy-coercion.md
```diff
@@ -0,0 +1,20 @@
...
+import { createEnv } from "@arkenv/nuxt";
+
+export const env = createEnv({
+  NUXT_PUBLIC_PORT: "number",
+  PORT: "number",
+});
+
+// Was "3000" (string) — now 3000 (number)
+env.NUXT_PUBLIC_PORT;
+env.NUXT_PUBLIC_PORT;
```
yamcodes:
```suggestion
```

packages/nuxt/src/create-env.ts
```diff
@@ -263,25 +263,35 @@ export function createEnvInternal(
...
 	);
 }
 
 /**
- * Wraps the validated environment object in a Proxy to enforce client/server security access rules.
+ * Wrap the validated environment object in a Proxy to enforce client/server security access rules.
+ *
+ * Schema-key reads always return the coerced validation target. Raw `useRuntimeConfig()` /
+ * `process.env` / `__NUXT__.config.public` strings are intentionally not preferred on get —
+ * those sources feed validation at create time, but serving them again would bypass coercion.
```
yamcodes, line 282: Interesting comment, is there a tradeoff here? Are we losing something with this PR?

yamcodes: can you give me a specific, case-by-case flow that would not work correctly after this PR lands and will be fixed in 1424 and how?

yamcodes: so is there a point to this PR at all if we're gonna ship #1424 anyway? isn't it superseded ?

PR #1422, opened 2026-07-20
feat: (v1) Auto-extend shared env in Nuxt strict-layout client entry

### LINKED ISSUE #1414: (v1) Auto-extend shared env in Nuxt strict-layout client entry

## Problem

In Nuxt strict layout, `env/client.ts` must explicitly import and extend `SharedSchema`, while `env/server.ts` now auto-extends the composed client env when `extends` is omitted. This leaves an asymmetric chain:

`SharedSchema` → client env → server env

The client-to-shared link still requires manual wiring and makes the canonical example harder to understand.

Related: #1307

## Desired behavior

When `@arkenv/nuxt/module` runs in strict layout, omitting `extends` from `@arkenv/nuxt/client` should automatically include the shared schema exported as `SharedSchema` from the strict layout's internal shared module. The canonical client entry should only declare client-specific variables. The server should continue to inherit both client and shared variables through its existing client-env auto-merge.

Any explicit `extends` property - including `extends: []`, a custom list, or `{ extends: undefined }` if the property is present - must be used as-is and opt out of the default shared merge, matching the server entry's current semantics. Explicit lists must not have `SharedSchema` appended implicitly.

Missing internal shared files, or a present shared file without a usable `SharedSchema` export, should fail with a clear strict-layout diagnostic rather than silently using an empty schema.

## Acceptance criteria

- [ ] The ArkType and Standard Schema client entries include shared variables at runtime and in their inferred return types when `extends` is omitted in strict layout.
- [ ] Explicit `extends` overrides auto-merge; `extends: []` includes no shared env; custom lists are not silently augmented with `SharedSchema`.
- [ ] Property presence of `extends` (including `{ extends: undefined }`) opts out, matching server-entry semantics.
- [ ] Missing shared file or missing usable `SharedSchema` export fails with an actionable strict-layout diagnostic.
- [ ] Existing server auto-extend behavior remains unchanged.
- [ ] Simple layout, legacy nested calls, and usage without the Nuxt module remain unchanged.
- [ ] Nuxt module wiring supports shared-env resolution across Vite, Nitro where relevant, and generated TypeScript configuration without leaking server-only schemas into browser bundles.
- [ ] Existing `NUXT_PUBLIC_` enforcement, runtimeConfig population, file watching, and client/server import blocking remain intact.
- [ ] CLI Nuxt strict-layout scaffolding omits the manual `SharedSchema` import and `extends` block from `env/client.ts` for ArkType, Zod, and Valibot templates; Next.js scaffolds remain unchanged.
- [ ] Strict-layout docs, examples, playgrounds, JSDoc, unit tests, type regressions, module-wiring tests, and build coverage reflect the simplified client entry and opt-out semantics.
- [ ] ArkType and Standard Schema entry points retain parity.

## Out of scope

- Next.js behavior.
- Changes to flat/simple layout.
- Changes to explicit `extends` semantics beyond matching the existing server opt-out model.
- File-based code generation for Nuxt.
- Restoring a public `@arkenv/nuxt/shared` export (removed by ADR 0014).
- Removing or changing `env/internal/shared.ts` as the strict-layout schema boundary.
- New module configuration options.

yamcodes:
Fixes #1414

## Summary
- Register `#arkenv/shared-schema` alias in `@arkenv/nuxt/module` for strict layout (alongside existing `#arkenv/client-env`)
- Auto-extend `SharedSchema` from `env/internal/shared.ts` in `@arkenv/nuxt/client` and `@arkenv/nuxt/standard/client` when `extends` is omitted
- Keep explicit `extends` (including `[]` and property presence) as an opt-out; fail clearly when shared file or `SharedSchema` export is missing
- Simplify Nuxt strict CLI scaffolds, docs, playground, and example client entries

## Test plan
- [x] `pnpm run typecheck`
- [x] `pnpm run test`
- [x] `pnpm run fix`
- [ ] Verify `apps/playgrounds/nuxt-strict` builds with simplified `env/client.ts`
- [ ] Confirm manual `extends: [SharedSchema]` still works as an opt-out


yamcodes:
Requires human review.

apps/www/content/docs/nuxt/layouts/strict.mdx
yamcodes, line 28: A little out of scope but (a) i don't like the way this bulletpoint is listed, it makes it feel like there are 4 pathways when there are 2, and (b) the note isn't rendering, if we want the note to have a title it should be ":::note[Migration]"""

yamcodes, line 44: This is very good just make sure that `type` isn't also exported from the Nuxt integration, nor any other repeated functions/consts/types. if there are any, lmk (dont change it here just inform)

In fact this question applies to any of the integrations not just nuxt

apps/www/content/docs/nuxt/layouts/strict.mdx
```diff
@@ -94,11 +79,14 @@ export const env = arkenv({
...
 ```
 
-When `@arkenv/nuxt/module` runs in strict layout, the server entry auto-merges client and shared variables. You do not need to import `./client` or pass `extends: [clientEnv]` unless you want an explicit override.
+When `@arkenv/nuxt/module` runs in strict layout, omitting `extends` auto-merges:
+
+- **Client entry** includes `SharedSchema` from `env/internal/shared.ts`
+- **Server entry** includes the composed client env (and therefore shared variables)
 
-The client entry still extends `SharedSchema` explicitly because automatic merging currently applies only to the server entry. The resulting chain is `SharedSchema` → client env → server env, so the server receives shared variables transitively through the client env.
+You do not need to import `SharedSchema` or `./client`, or pass `extends`, unless you want an explicit override.
```
yamcodes: This sentence feels unnecessary, new users don't know the history of the code so they don't need "you do not need" style sentences. this feels very meta-decision-esque.

might apply elsewhere.

apps/www/content/docs/nuxt/layouts/strict.mdx
```diff
@@ -94,11 +79,14 @@ export const env = arkenv({
 });
 ```
 
-When `@arkenv/nuxt/module` runs in strict layout, the server entry auto-merges client and shared variables. You do not need to import `./client` or pass `extends: [clientEnv]` unless you want an explicit override.
+When `@arkenv/nuxt/module` runs in strict layout, omitting `extends` auto-merges:
```
yamcodes: 1. of course we're talking about strict layout
2. `extends` doesnt need to be mentioned at all in the "Your schema" section

I think let's just move it to its own "Extending your schema" / "Exposing client variables" section, take a look at the flat layout page

apps/www/content/docs/nuxt/layouts/strict.mdx
```diff
@@ -94,11 +79,14 @@ export const env = arkenv({
...
+
+- **Client entry** includes `SharedSchema` from `env/internal/shared.ts`
+- **Server entry** includes the composed client env (and therefore shared variables)
 
-The client entry still extends `SharedSchema` explicitly because automatic merging currently applies only to the server entry. The resulting chain is `SharedSchema` → client env → server env, so the server receives shared variables transitively through the client env.
+You do not need to import `SharedSchema` or `./client`, or pass `extends`, unless you want an explicit override.
 
-Auto-merge only applies when the `extends` key is **omitted**. Passing `extends` at all - including `extends: []` or a list that does not include `clientEnv` - opts out and uses your list as-is.
+Auto-merge only applies when the `extends` key is **omitted**. Passing `extends` at all - including `extends: []` or a custom list - opts out and uses your list as-is. Explicit lists are not silently augmented with `SharedSchema` or `clientEnv`.
```
yamcodes: this is confusing as is, and also i'd appreciate a mention (perhaps in our new section) on how to customize the behavior of extending the shared vars - is it possible to change the location from our default env/internal/shared.ts to something else like env/hidden/shared.ts if the client wanted to ? is it via `withArkEnv`-style function or in some config? it'd also be nice to add an "Configuration & API" page just like Next has - take a look

packages/nuxt/tsdown.config.ts
yamcodes, line 1: will the user ever see those virtual packages, need to interact with, or customize them ?

in their own code ?

apps/www/content/docs/nuxt/layouts/strict.mdx
yamcodes: Addressed in 3bab13ca — collapsed the import mental model into two pathways (flat vs strict), with the shared/standalone import notes nested under strict, and fixed the callout to `:::note[Migration]`.

yamcodes: Informational (no code change here): `type` **is** re-exported from the Nuxt integration today — `@arkenv/nuxt`, `@arkenv/nuxt/client`, and `@arkenv/nuxt/server` all do `export { type } from "@arkenv/core"`. Next.js does the same from `@arkenv/nextjs`, `/client`, `/server`, and `/react-server` (and codegen templates re-export it too). Vite/Bun plugins do not appear to re-export `type`. Happy to open a follow-up if we want to stop re-exporting and make `@arkenv/core` the sole source for `type` across integrations.

apps/www/content/docs/nuxt/layouts/strict.mdx
yamcodes: Addressed in 3bab13ca — removed the “you do not need…” sentence and similar meta wording from the schema section / manual setup.

apps/www/content/docs/nuxt/layouts/strict.mdx
yamcodes: Addressed in 3bab13ca — “Your schema” now only shows the three files. Composition / `extends` lives under a new **Extending your schema** subsection (modeled after flat’s “Exposing client variables”).

apps/www/content/docs/nuxt/layouts/strict.mdx
yamcodes: Addressed in 3bab13ca:

- Clarified in **Extending your schema** that `internal/shared.ts` is fixed relative to the schema directory — there is no option to relocate it to e.g. `env/hidden/shared.ts`. Custom composition is via explicit `extends` in the client/server entries (not a `withArkEnv`-style wrapper; Nuxt uses the module + `arkenv` config key).
- Added a Nuxt [Configuration & API](https://arkenv.js.org/docs/nuxt/configuration) page (mirroring Next’s), covering `schemaPath`, `layout`, `validate`, logging, and the strict composition rules.

packages/nuxt/tsdown.config.ts
yamcodes: No — users should never import or customize `#arkenv/client-env` / `#arkenv/shared-schema` in their own code. Those are internal virtual specifiers the module registers so the package entries can statically import a fallback outside strict layout and resolve to the project files when the module is active. App code keeps importing from `@arkenv/nuxt/client` / `/server` (and `./internal/shared` only when explicitly opting into custom `extends`).

apps/www/content/docs/nuxt/layouts/strict.mdx
yamcodes: we dont need it anymore, users have to install arkenv core anyway right? so they can just use it from there.
please make a ticket for it unless you have any strong pushbacks here based on our docs ?

and make sure nothing else gets re-exported , thanks.

yamcodes: No pushback from docs — they already use `import { type } from \"@arkenv/core\"` (and don't show importing `type` from the Nuxt/Next packages).

Filed https://github.com/yamcodes/arkenv/issues/1437.

Also found the same pattern for `Infer` (`export type { Infer } from \"@arkenv/core\"` on Nuxt flat entry + Next.js index/react-server) and Next codegen templates that re-export `type` from the integration — included those in the ticket so we strip the whole redundant surface, not just `type`.

.changeset/nuxt-strict-client-shared-auto-extend.md
```diff
@@ -0,0 +1,22 @@
...
+
+```ts
+import arkenv from "@arkenv/nuxt/client";
+
+export const env = arkenv({
+	NUXT_PUBLIC_API_URL: "string",
+});
+```
+
+Auto-merge only runs when the `extends` key is omitted. Any explicit `extends` - including `extends: []` or a custom list - is used as-is and opts out of auto-merge. Missing `env/internal/shared.ts` or a missing usable `SharedSchema` export fails with a clear strict-layout diagnostic.
```
yamcodes: "Missing `env/internal/shared.ts` or a missing usable `SharedSchema` export fails with a clear strict-layout diagnostic."

Thats weird so a simple example of just a single env in client and a single env in server (nothing shared) will fail out of box?

yamcodes: No — client+server-only still works out of the box. Strict layout always includes `env/internal/shared.ts` as part of the file layout, but `SharedSchema` itself can be empty (`type({})`) when you have no shared variables. What fails is a *missing file* or *missing/unusable export* (so we don't silently pretend shared is empty).

Clarified in the changeset + Nuxt strict docs in 599a7d41.

yamcodes:
LGTM

Issue #1105, opened 2026-05-31
(v1) Adopt the env object as the canonical surface for Vite/Bun (transform design)


yamcodes:
> **Design spike resolved.** This issue originally asked what the Vite/Bun plugins should learn from the Next.js work (#1084, #1092). The answer, after a full adversarial design review, is bigger than a pattern port: **the imported `env` object becomes the canonical surface for all integrations, implemented for Vite/Bun via the transform design.** The decision and its full argument tree are recorded in **ADR 0015** (#1326) on `dev`; on `v1` the same decision is **ADR 0021**. The July 5 agent brief (Next-shaped codegen port) and the July 15 landscape comment (keep the plugin-env/codegen seam) are both superseded by this decision.

## Decision

`import { env } from "./env"` is the **one** way to consume validated env across Next, Nuxt, Vite, and Bun — on both sides of the network boundary. For Vite/Bun this is **SvelteKit-shaped, not Next-shaped**:

- **`env.ts` is the single typed source of truth.** The user writes `createEnv`; types flow from normal inference. No `env.gen.ts`, no committed codegen files, no `runtimeEnv` destructuring.
- **Server graph executes `env.ts` as-is** → boot-time validation and coercion against the real deployment environment (the fail-fast guarantee fullstack Vite/Bun apps currently lack entirely).
- **Client graph gets a transformed module**: the plugin rewrites the env module in client bundles — client-prefixed keys become inlined, build-validated, coerced literals; the validator import is stripped; server-only keys become a small throwing guard. No validator ships to the browser; nothing re-validates client-side.
- **Strict layout** reuses the ADR 0013 client-import blocker (already a Vite plugin); Bun gets the equivalent via resolver hooks.
- **Plumbing** consolidates in `@arkenv/build`; `clientPrefix` becomes first-class plugin config for all frameworks (per ADR 0014's note).
- **Schema/define path dropped on v1 (Option 3 — #1333).** No escape hatch, no “SPA mode,” no co-equal ambient surface. `arkenv(schema)` + native accessors + `.d.ts` remains on **v0 (`dev`) only** so stable production apps are not broken; v1 removes it completely. Docs/CLI/examples/skill teach the env object exclusively. ADR 0015/0021’s “SPA-mode permanence” clause is overturned (amendment via #1333).

## Landing strategy

[...]

PR #1423, opened 2026-07-20
feat: (v1) env-module transform — canonical env object surface

### LINKED ISSUE #1328: (v1) feat(vite-plugin): env-module transform — canonical env object surface
## Summary

Implement the ADR 0015 transform design in `@arkenv/vite-plugin`: the user's `env.ts` (calling `createEnv`) becomes the canonical typed surface, executed as-is in the server graph and rewritten in the client graph.

## Scope

- **Env-module identification**: convention/config for locating the env module (like ADR 0013's `schemaPath`), robust to aliased imports and monorepo paths.
- **Client-graph transform** (discriminated via the `ssr` flag in transform hooks): replace client-prefixed keys with inlined, build-validated, **coerced** literals; strip the validator import; emit a small throwing guard for server-only keys ("trust the proxy" per ADR 0010). No validator ships to the browser; no client-side re-validation.
- **Server-graph passthrough**: `env.ts` executes normally → boot-time validation against the real deployment environment.
- **`clientPrefix` as first-class plugin config** (per ADR 0014's note), defaulting to Vite's `envPrefix`.
- **HMR invalidation** when the env module or `.env` files change.

## Out of scope

- Strict layout / import blocking (#1330).
- `@arkenv/build` plumbing consolidation (#1331).
- Any `env.gen.ts`-style codegen or `runtimeEnv` option (explicitly rejected — see ADR 0015 direction 2).

## Acceptance criteria

- [ ] `import { env } from "./env"` works identically in client and SSR code in a fullstack Vite app (verify against the `with-solid-start` example extended with a server-only key).
- [ ] Client chunks contain inlined coerced literals + guard stub only — no validator code.
- [ ] Reading a server-only key on the client throws.
- [ ] Existing plugin + `.d.ts` path keeps working (SPA mode).

## Landing

**v1 branch only** — see ADR 0015 landing strategy. Part of #1105 (ADR: #1326).

yamcodes:
Fixes #1328

## Summary
- Add Vite plugin **transform mode** (`arkenv()` / `arkenv({ schemaPath, clientPrefix })`): rewrite the user's `env.ts` in the client graph with build-validated coerced literals + server-key throwing guards; leave the SSR graph untouched for boot-time validation (ADR 0015).
- Keep SPA mode (`arkenv(schema)` + `ImportMetaEnvAugmented` / `define`) unchanged.
- Extend `with-solid-start` (and playground) with `src/env.ts` including a server-only `DATABASE_URL`.

## Test plan
- [x] `pnpm --filter @arkenv/vite-plugin typecheck`
- [x] `pnpm --filter @arkenv/vite-plugin exec vitest run` (SPA + transform fixtures)
- [ ] Client chunk has no `@arkenv/core` / `arktype` validator code
- [ ] Reading `env.DATABASE_URL` on the client throws; SSR can read it
- [ ] Existing SPA `define` path still works


yamcodes:
I like this, but it requires human review.

yamcodes:
Addressed the SolidStart README regression:

- Updated `apps/playgrounds/solid-start/README.md` to document transform mode (`src/env.ts` + `arkenvVitePlugin()`)
- Synced the same README into `examples/with-solid-start/README.md`
- Fixed `env.ts` JSDoc to say "SolidStart example" (not "playground")

.changeset/vite-env-module-transform.md
```diff
@@ -0,0 +1,30 @@
+---
+"@arkenv/vite-plugin": minor
+---
+
+#### Add env-module transform for fullstack Vite apps
+
+Enable `import { env } from "./env"` as the canonical surface: the plugin rewrites `env.ts` in the client graph (inlined coerced literals + server-key guards) and leaves it untouched in the SSR graph for boot-time validation.
```
yamcodes: "rewrites `env.ts`" is not clear enough

.changeset/vite-env-module-transform.md
```diff
@@ -0,0 +1,30 @@
...
+
+```ts
+// vite.config.ts / app.config.ts
+import arkenv from "@arkenv/vite-plugin";
+
+export default {
+  plugins: [
+    arkenv(), // or arkenv({ schemaPath: "src/env.ts", clientPrefix: "VITE_" })
```
yamcodes: it's not clear enough what an empty `arkenv()` does here

.changeset/vite-env-module-transform.md
```diff
@@ -0,0 +1,30 @@
...
+// src/env.ts
+import arkenv from "@arkenv/core";
+
+export const env = arkenv({
+  DATABASE_URL: "string",
+  VITE_API_URL: "string",
+});
+```
+
+SPA mode (`arkenv(schema)` + `ImportMetaEnvAugmented` for `import.meta.env`) remains supported.
```
yamcodes: SPA mode is a new concept introduced here, needs to be more clear

It's also not clear if this change only supports arkenv core or also arkenv standard

.changeset/vite-env-module-transform.md
```diff
@@ -0,0 +1,30 @@
+---
+"@arkenv/vite-plugin": minor
```
yamcodes, line 2: are we sure it's not a breaking change?

yamcodes: if not - was it smart to introduce this change in `v1` branch or could we have already introduced it in v0 (`dev`) ?

apps/playgrounds/solid-start/src/app.tsx
```diff
@@ -1,8 +1,13 @@
 import { createSignal } from "solid-js";
+import { env } from "./env";
 import "./app.css";
 
+const serverDatabaseUrl =
+	typeof window === "undefined" ? env.DATABASE_URL : null;
```
yamcodes: is this the best way in solid start to do this? using a `window === "undefined"` check ?

apps/playgrounds/solid-start/src/global.d.ts
```diff
@@ -1,10 +1,4 @@
...
-type ImportMetaEnvAugmented =
-	import("@arkenv/vite-plugin").ImportMetaEnvAugmented<
-		typeof import("../app.config").Env
-	>;
-
-// Augment import.meta.env with your schema
-// Only `VITE_*` prefixed variables will be included
-interface ImportMetaEnv extends ImportMetaEnvAugmented {}
+// Fullstack apps use `import { env } from "./env"` (transform mode).
+// SPA mode can still augment import.meta.env via ImportMetaEnvAugmented.
```
yamcodes: Why? why does SPA mode need a special distinction? If it's just "for continuity with the feature we had" this is not a good justification. if it's "to make migration to ArkEnv from other solutions or custom/no validation" it's an ok  justificaiton but then i wouldn't brand it as "SPA mode", i'd brand it as an escape hatch or just not offer it altogether (maybe i'd offer an ArkEnv CLI migration instead)

apps/playgrounds/solid-start/README.md
```diff
@@ -1,76 +1,47 @@
...
-This example demonstrates how to use [@arkenv/vite-plugin](https://arkenv.js.org/docs/vite-plugin) with [SolidStart](https://start.solidjs.com). It showcases:
+This example demonstrates `@arkenv/vite-plugin` **transform mode** with [SolidStart](https://start.solidjs.com):
 
-- **Environment variable validation** at build-time with ArkEnv
-- **Typesafe `import.meta.env`** with full TypeScript support
-- **Client-side environment variables** with automatic filtering of `VITE_*` prefixed variables
+- A single `src/env.ts` is the typed source of truth (`import { env } from "./env"`)
+- **Client graph**: plugin inlines coerced `VITE_*` literals and guards server-only keys
+- **SSR graph**: `env.ts` runs as-is → boot-time validation against the real environment
+- Reading `env.DATABASE_URL` in the browser throws (trust-the-proxy)
```
yamcodes: trust-the-proxy? whats that?

packages/vite-plugin/src/env-module.ts
yamcodes, line 1: this file is starting to become a bit long, consider to `/modularize`

packages/vite-plugin/src/vite-plugin-generic.ts
yamcodes, line 1: this file is starting to become a bit long, consider to `/modularize`

packages/vite-plugin/tsdown.config.ts
```diff
@@ -7,6 +7,6 @@ export default defineConfig({
 	fixedExtension: false,
 	deps: {
 		alwaysBundle: ["@repo/log", "@repo/types", "@repo/utils"],
-		neverBundle: ["vite"],
+		neverBundle: ["vite", "@arkenv/build", "jiti"],
```
yamcodes, line 13: why we added jiti here

.changeset/vite-env-module-transform.md
yamcodes: Clarified in the changeset: the plugin replaces the env module in the browser build with coerced literals + throwing getters for server-only keys, and leaves it alone on SSR.

.changeset/vite-env-module-transform.md
yamcodes: Clarified: `arkenv()` with no schema auto-discovers `src/env.ts` / `env.ts` (or take `schemaPath` / `clientPrefix`).

.changeset/vite-env-module-transform.md
yamcodes: Dropped the "SPA mode" branding in the changeset. Existing `arkenv(schema)` + `import.meta.env` define path is unchanged. Transform works for both `@arkenv/vite-plugin` (core) and `@arkenv/vite-plugin/standard`.

.changeset/vite-env-module-transform.md
yamcodes: Not a breaking change — existing `arkenv(schema)` behavior is unchanged; `arkenv()` is an additive overload (hence `minor`).

Landing on `v1` only matches ADR 0015's landing strategy: the transform design is meant to ship with the v1 defaults/migration story rather than soft-landing on `dev` first and re-implementing on v1.

apps/playgrounds/solid-start/src/app.tsx
yamcodes: Switched to Solid's idiomatic `isServer` from `solid-js/web`.

apps/playgrounds/solid-start/src/global.d.ts
yamcodes: Agreed the dual branding was wrong for the example. Removed SPA-mode callouts from the SolidStart README / `global.d.ts`.

Keeping the existing `arkenv(schema)` + `define` API as a backward-compat path (issue acceptance + ADR 0015) so current Vite users don't need a forced migration — not as a co-equal "mode" in docs. Happy to revisit whether we deprecate that path later.

apps/playgrounds/solid-start/README.md
yamcodes: Removed the jargon. Now just says reading `env.DATABASE_URL` in the browser throws.

packages/vite-plugin/src/env-module.ts
yamcodes: Modularized into `transform-options`, `env-module-path`, `classify-env-keys`, `load-validated-env`, `generate-client-env-module`, with `env-module.ts` as a barrel.

packages/vite-plugin/src/vite-plugin-generic.ts
yamcodes: Split into `define-plugin.ts` (schema/`define` path) and `transform-plugin.ts`, with `vite-plugin-generic.ts` as the thin orchestrator.

packages/vite-plugin/tsdown.config.ts
yamcodes: Jiti loads the user's `env.ts` at build time so we can get coerced values for client inlining. Listed in `neverBundle` so it stays an external runtime dep (Node resolves the project's modules) rather than being bundled into the plugin.

.changeset/vite-env-module-transform.md
```diff
@@ -0,0 +1,38 @@
...
+---
+
+#### Add env-module transform for fullstack Vite apps
+
+Make `import { env } from "./env"` the recommended surface for Vite apps that have both client and server code.
+
+With no schema argument, the plugin finds `src/env.ts` / `env.ts` (or `schemaPath`) and:
+
+- **Browser build**: replaces that module with an object of build-validated, coerced literals for client-prefixed keys, plus getters that throw for server-only keys — no `@arkenv/core` / validator code in the client bundle
+- **SSR / server build**: leaves the module alone so `createEnv`/`arkenv` still validates against the real deployment environment at boot
```
yamcodes: i dont like the way this is written, "make X the recommended surface" but that surface is new to the PR, it didnt exist before. you're introducing something not recommending something that exists already.

also, "with no augment, ..." why talk about schema augments if nobody asked about it?

i think you need to take a look at `/changeset` skill.

it's also not clear in the Browser build section - does code actually get changed for the user?  is their schema getting changed ?

some of this changeset feels not consumer friendly and too impl. detail oriented / meta-decision oriented

yamcodes: Rewrote the changeset to introduce the feature (not "recommend" an existing surface), drop the schema-argument digression, and clarify that the bundler inlines client values at build time without rewriting `env.ts` on disk.

examples/basic/package.json
```diff
@@ -9,7 +9,7 @@
 		"clean": "rimraf dist node_modules"
 	},
 	"dependencies": {
-		"@arkenv/core": "1.0.0-alpha.7",
+		"@arkenv/core": "1.0.0-alpha.4",
```
yamcodes, line 12: why this decrease from 7 to 4, whats going on here

yamcodes: This is from fixing the examples sync package map in `scripts/sync-lib/workspace.js`.

Previously `@arkenv/core` was incorrectly resolved from `packages/arkenv` (the **CLI** package, whose `name` is `arkenv` and is currently at `1.0.0-alpha.9`). That made examples pick up the CLI’s version number (`alpha.7` / `alpha.8`) even though that is not a published `@arkenv/core` release.

`@arkenv/core` lives in `packages/core` and is actually at **`1.0.0-alpha.4`** (matches npm). After correcting the map and re-syncing, examples pin that real version.

So this is not a downgrade of core — it stops examples from inventing a non-existent `@arkenv/core@1.0.0-alpha.7`.

apps/playgrounds/solid-start/src/global.d.ts
yamcodes: not a good enough justification :( unresolving. we need to align on this in the parent issue(s). let's discuss

yamcodes:
soft-approving this but i want to understand what we do with the "SPA mode", not sure we need it but it also feels out of scope.

apps/playgrounds/solid-start/src/global.d.ts
yamcodes: Aligned offline (/grill-with-docs). Summary for the thread + parent issues:

**In scope for this PR (ship as-is)**
- Env-module transform is the canonical Vite surface (`import { env } from "./env"`).
- Schema/`define` + `ImportMetaEnvAugmented` **keeps working** (#1328 acceptance) — continuity-alone is *not* the product story, and we are **not** branding this as “SPA mode” in examples/docs here.
- Settling lasting fate of that path is **not a merge blocker**.

**Deferred to #1105 / ADR 0015 follow-up** (needs hands-on play first)
- Escape hatch forever vs time-boxed deprecate/remove vs hard removal.
- Current lean for defaults: teach env object only in docs/CLI; leave schema/define unbranded until we pick.

Resolving this thread so #1423 can merge. Product decision continues on the parent issue.

PR #1459, opened 2026-07-22
feat: (v1) env-module transform — canonical env object surface

### LINKED ISSUE #1329: (v1) feat(bun-plugin): env-module transform — canonical env object surface
## Summary

Implement the ADR 0015 transform design in `@arkenv/bun-plugin`: the user's `env.ts` is the canonical typed surface — transformed in client bundles, executed live on the server.

## Scope

- **Client bundles (`Bun.build`)**: per-target `onLoad` transform of the env module — client-prefixed keys (`BUN_PUBLIC_*` via first-class `clientPrefix` config, per ADR 0014) become inlined, build-validated, **coerced** literals; validator import stripped; throwing guard emitted for server-only keys. No validator ships; no client re-validation.
- **Server runtime (`bun run` / `Bun.serve`)**: `env.ts` executes as-is → boot-time validation and coercion against the real deployment environment. The existing zero-config discovery (`src/env.ts` / `env.ts` via `bunfig.toml`, ADR 0001) is reused for module identification.
- **Env-module identification** shared conventions with the Vite implementation (#1328).

## Out of scope

- Strict layout / import blocking (#1330).
- `@arkenv/build` plumbing consolidation (#1331).
- Any `env.gen.ts`-style codegen or `runtimeEnv` option (rejected — ADR 0015 direction 2).

## Acceptance criteria

- [ ] One `import { env } from "./env"` serves both a `Bun.build` client bundle and `Bun.serve` server code (extend the `with-bun-react` example with a server-only key).
- [ ] Client bundles contain inlined coerced literals + guard stub only.
- [ ] Server boot fails fast on missing/invalid env with real deploy values.
- [ ] Existing plugin + `bun-env.d.ts` path keeps working (SPA mode).

## Landing

**v1 branch only** — see ADR 0015 landing strategy. Part of #1105 (ADR: #1326).

yamcodes:
Fixes #1329

## Summary
- Add dual-mode `@arkenv/bun-plugin`: `arkenv()` / zero-config default rewrites `env.ts` in browser bundles (inlined coerced `BUN_PUBLIC_*` literals + server-key guards); `arkenv(schema)` keeps SPA `process.env` rewriting
- Server/`Bun.serve` continues to execute `env.ts` as-is for boot-time validation against the real environment (plugin is `target: "browser"`)
- Extend `with-bun-react` with a server-only `DATABASE_URL` and shared `import { env } from "./env"` on client and server

## Test plan
- [ ] `pnpm --filter @arkenv/bun-plugin test` passes (helpers + transform onLoad + SPA regression)
- [ ] Client `Bun.build` bundle contains inlined literals / guard stub and no validator imports
- [ ] `bun run` / `Bun.serve` fails fast when required env is missing
- [ ] SPA `arkenv(schema)` + `ProcessEnvAugmented` path still works

### DIFF
--- a/apps/playgrounds/bun-react/bin/build.ts
+++ b/apps/playgrounds/bun-react/bin/build.ts
@@ -1,11 +1,11 @@
-import { arkenv } from "@arkenv/bun-plugin";
+import arkenv from "@arkenv/bun-plugin";
 
 const result = await Bun.build({
 	entrypoints: ["./src/index.html"],
 	outdir: "./dist",
 	target: "browser",
 	minify: true,
-	plugins: [arkenv()],
+	plugins: [arkenv],
 });
 
 if (!result.success) {
--- a/apps/playgrounds/bun-react/bin/build.ts
+++ b/apps/playgrounds/bun-react/bin/build.ts
@@ -1,11 +1,11 @@
-import arkenv from "@arkenv/bun-plugin";
+import { arkenv } from "@arkenv/bun-plugin";
 
 const result = await Bun.build({
 	entrypoints: ["./src/index.html"],
 	outdir: "./dist",
 	target: "browser",
 	minify: true,
-	plugins: [arkenv],
+	plugins: [arkenv()],
 });
 
 if (!result.success) {
--- a/apps/playgrounds/bun-react/src/app.tsx
+++ b/apps/playgrounds/bun-react/src/app.tsx
@@ -1,5 +1,5 @@
-import { APITester } from "./api-tester";
 import { env } from "@/env";
+import { APITester } from "./api-tester";
 import "./index.css";
 
 import logo from "./logo.svg";
--- a/apps/playgrounds/bun-react/src/app.tsx
+++ b/apps/playgrounds/bun-react/src/app.tsx
@@ -1,5 +1,5 @@
 import { APITester } from "./api-tester";
-import { env } from "./env";
+import { env } from "@/env";
 import "./index.css";
 
 import logo from "./logo.svg";
--- a/apps/playgrounds/bun-react/src/app.tsx
+++ b/apps/playgrounds/bun-react/src/app.tsx
@@ -1,4 +1,5 @@
 import { APITester } from "./api-tester";
+import { env } from "./env";
 import "./index.css";
 
 import logo from "./logo.svg";
@@ -28,18 +29,18 @@ export function App() {
 				<tbody>
 					<tr>
 						<td>BUN_PUBLIC_API_URL</td>
-						<td>{process.env.BUN_PUBLIC_API_URL}</td>
-						<td>{typeof process.env.BUN_PUBLIC_API_URL}</td>
+						<td>{env.BUN_PUBLIC_API_URL}</td>
+						<td>{typeof env.BUN_PUBLIC_API_URL}</td>
 					</tr>
 					<tr>
 						<td>BUN_PUBLIC_DEBUG</td>
-						<td>{String(process.env.BUN_PUBLIC_DEBUG)}</td>
-						<td>{typeof process.env.BUN_PUBLIC_DEBUG}</td>
+						<td>{String(env.BUN_PUBLIC_DEBUG)}</td>
+						<td>{typeof env.BUN_PUBLIC_DEBUG}</td>
 					</tr>
 				</tbody>
 			</table>
 			{/* Print whether we are in "build" or in the dev server */}
-			<p>Mode: {process.env.NODE_ENV}</p>
+			<p>Mode: {env.NODE_ENV}</p>
 		</div>
 	);
 }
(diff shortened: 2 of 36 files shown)


packages/bun-plugin/src/bun-plugin-generic.ts
```diff
@@ -1,111 +1,82 @@
...
-		return {
-			name: pluginName,
-			setup(build) {
-				registerLoader(build, envMap);
-			},
-		} satisfies BunPlugin;
+		return createDefinePlugin(
+			coreArkenv,
+			pluginName,
+			schemaOrOptions as CompiledEnvSchema | SchemaShape,
```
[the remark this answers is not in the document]
yamcodes: Addressed in e1ff68a4 — cast to `Record<string, unknown>` so the SPA define path typechecks.

Also fixed the related CI failures in the same commit:
- Updated `apps/playgrounds/bun-react` (source of truth) to the transform `env` surface so `check-sync` and `bun-react-playground#build` pass
- Documented the Bun.serve env-refresh limitation in `transform-plugin.ts` (`onStart` only revalidates on rebuild; no Vite-style HMR hook)

yamcodes:
Addressed review + CI failures in e1ff68a4:

| Issue | Action |
| --- | --- |
| `createDefinePlugin` schema type mismatch (typecheck) | Cast to `Record<string, unknown>` |
| `check-sync` / out-of-sync `with-bun-react` | Updated playground source of truth (`apps/playgrounds/bun-react`) |
| `bun-react-playground#build` (hybrid expected `env` object) | Migrated playground to transform `env.ts` + `arkenv()` |
| Dev-server env refresh gap | Documented Bun limitation in transform-plugin `@remarks` (restart/rebuild required) |

apps/playgrounds/bun-react/bin/build.ts
```diff
@@ -1,11 +1,11 @@
-import arkenv from "@arkenv/bun-plugin";
+import { arkenv } from "@arkenv/bun-plugin";
 
 const result = await Bun.build({
 	entrypoints: ["./src/index.html"],
 	outdir: "./dist",
 	target: "browser",
 	minify: true,
-	plugins: [arkenv],
+	plugins: [arkenv()],
```
yamcodes: justify why this needs to be a function now

apps/playgrounds/bun-react/bin/build.ts
```diff
@@ -1,11 +1,11 @@
-import arkenv from "@arkenv/bun-plugin";
+import { arkenv } from "@arkenv/bun-plugin";
```
yamcodes: why not a default export? in `@arkenv/core` we use default export

apps/playgrounds/bun-react/src/app.tsx
```diff
@@ -1,4 +1,5 @@
 import { APITester } from "./api-tester";
+import { env } from "./env";
```
yamcodes: consider `~/env` or `@/env` absolute import alias, whatever the best practice is for Bun (scour their docs)

apps/playgrounds/bun-react/src/env.ts
yamcodes, line 1: i feel like thats a bit too impl. detail heavy. most consumers wont care. i know you used the vite example but i'd fix it there too.

apps/playgrounds/bun-react/bin/build.ts
yamcodes: It doesn't — reverted to the default-export plugin object (`plugins: [arkenv]`). The hybrid default already carries transform `setup`, so calling `arkenv()` was unnecessary for zero-config.

apps/playgrounds/bun-react/bin/build.ts
yamcodes: Agreed — switched back to `import arkenv from \"@arkenv/bun-plugin\"` to match `@arkenv/core` / the previous example style.

apps/playgrounds/bun-react/src/app.tsx
yamcodes: Bun rewrites imports via `tsconfig.json` `paths` ([docs](https://bun.com/docs/guides/runtime/tsconfig-paths)). This playground already maps `@/*` → `./src/*`, so switched to `import { env } from \"@/env\"`.

apps/playgrounds/bun-react/src/env.ts
yamcodes: Slimmed to \`/** Validated environment for this app. */\` here and in the SolidStart playground/example too.

yamcodes:
Addressed the example review notes in 73ede2dd:

| Comment | Action |
| --- | --- |
| Why `arkenv()`? | Reverted to default-export plugin object — hybrid already has transform setup |
| Prefer default export | `import arkenv from "@arkenv/bun-plugin"` |
| Absolute env import | `import { env } from "@/env"` (existing `@/*` tsconfig paths; Bun-supported) |
| Heavy `env.ts` JSDoc | Slimmed for bun-react + solid-start |

PR #1566, opened 2026-08-12
fix(www): prevent hero fade from playing twice after hydrate

yamcodes:
## Summary
- The homepage hero fade played twice because `HeroMotion` hid already-visible SSR content then faded it in, and fumadocs `HomeLayout` (a client boundary) remounted `.rise` nodes after hydrate, restarting the CSS animation.
- Match [open-slide](https://github.com/1weiho/open-slide): CSS-only rise with `animation-fill-mode: backwards` so first paint is already hidden, and keep the hero out of `HomeLayout` (server `<main>` + `SiteNavHome`).
- Drop `HeroMotion` and its unit suite; add a Playwright assertion that each intro `.rise` / `.rise-blur` fires `animationstart` once.

## Test plan
- [ ] Load `/` and confirm the hero rises once (no second fade / flash)
- [ ] Reload a few times in both `pnpm www` (Strict Mode) and a production build
- [ ] Confirm `prefers-reduced-motion: reduce` still skips the entrance
- [ ] `pnpm exec playwright test tests/homepage.test.ts --project=chromium -g "hero intro rise"` from `apps/playwright-www`


yamcodes:
Addressed the redundant-selector nit in f681c833.

Dropped the unstamped `[data-hero-motion=play] .rise` / `.rise-blur` rules from the reduced-motion block — those no longer animate.

Kept the stamped `[data-hero-run]` pair: they match the play animation’s specificity so a live `prefers-reduced-motion` toggle still cancels an in-flight rise (the effect does not re-run). The generic `.home-aurora .rise` rules are too weak for that case.

yamcodes:
LGTM

yamcodes:
Addressed the latest review notes:

- Updated this PR description and test plan to the open-slide CSS-only rise (no `data-hero-run` / `hero-motion.test.tsx`).
- Added a Playwright check that each intro `.rise` / `.rise-blur` fires `animationstart` once after the entrance settles (`61cfe1c7`).

PR #1569, opened 2026-08-12
feat(standard): optional `toJsonSchema` coercion escape hatch (#1564)

### LINKED ISSUE #1564: (v1) options.toJsonSchema: Valibot-friendly coercion escape hatch + Validators guides
### Summary

ArkEnv coercion for `@arkenv/standard` introspects **Standard JSON Schema** per key. Zod embeds that on the schema; Valibot intentionally does not ([valibot.dev/guides/json-schema](https://valibot.dev/guides/json-schema/)) — conversion lives in `@valibot/to-json-schema`. Today Valibot users either wrap every key with `toStandardJsonSchema` (ugly) or skip ArkEnv coercion and use pipes/transforms.

Add an optional **`toJsonSchema`** config callback as a vendor-neutral escape hatch, and document it with a new **Guides → Validators** section (ArkType, Zod, Valibot).

### Current behavior

- `extractJsonSchema` only succeeds when a field already exposes JSON Schema (`~standard.jsonSchema`, etc.).
- Built-in probes swallow converter throws and treat that key as missing (best-effort **pre-coercion**). Unchanged for probes.
- Raw Valibot fields land in `missingKeys`; ArkEnv does not coerce them.
- Docs describe coercion as requiring Standard JSON Schema on the value itself.
- `examples/with-valibot` uses `v.transform(Number)` instead of ArkEnv coercion.

### Desired behavior

**Runtime**

- Add `toJsonSchema` to `ParseStandardConfig` / `@arkenv/standard` options:

```ts
toJsonSchema?: (schema: StandardSchemaV1) => object | undefined;
```

- Call it **only as a fallback** after existing Standard JSON Schema probes fail (hybrids: Zod + Valibot in one schema work).
- Not invoked when the option is omitted, when `coerce` is `false`, or when a built-in probe already produced a schema for that key.
- **Return `undefined`:** explicit skip for that key (same as a failed probe / `missingKeys`).
- **Throw:** wrap in `ArkEnvError` for that key (config/schema error, like invalid-schema), fail the parse. Do not swallow. Do not continue as “missing JSON Schema.”
- **Non-plain object** (`Date`, array, function, boxed primitive, etc.): same as throw — `ArkEnvError` for that key, fail the parse. Check with `Object.prototype.toString.call(val) === "[object Object]"`. Do **not** reuse `assertStandardSchemaMap`’s `typeof === "object" && !Array.isArray` test (that accepts `Date`). There is no existing `isPlainObject` helper; add a small internal one for this path. Do not export it. Do not change schema-map guards in this issue.
- Truthy plain object → use as that key’s JSON Schema for pre-coercion.
- No new peers on `@arkenv/standard`. Users who need the hatch install `@valibot/to-json-schema` themselves.
- Zod-only users omit the option; unchanged.
- Do not add this field to ArkType `ArkEnvConfig`. Vite/Bun `/standard` configs that already alias `ParseStandardConfig` inherit it; do not invent a second plugin-only API.

Documented Valibot wiring (explicit wrapper — **not** a bare `{ toJsonSchema }` function reference):

```ts
import arkenv from "@arkenv/standard";
import * as v from "valibot";
import { toJsonSchema } from "@valibot/to-json-schema";

export const env = arkenv(
  { PORT: v.number(), DEBUG: v.boolean() },
  {
    toJsonSchema: (schema) =>
      toJsonSchema(schema as v.GenericSchema, {
        typeMode: "input",
        target: "draft-07",
      }),
  },
);
```

`target: "draft-07"` matches built-in probes. `typeMode: "input"` coerces env strings toward the schema’s input type. Bare `toJsonSchema` would use Valibot’s default `typeMode: "ignore"` and silently mis-coerce piped schemas — that shorthand is rejected. Prefer `v.number()` / `v.boolean()` in the example (no `v.transform(Number)`). The wrapper needs `as v.GenericSchema` on the input; the `object` return needs no assertion.

**Docs (wider than the option alone)**

- New **Guides → Validators** section with full cookbooks: **ArkType**, **Zod**, **Valibot**.
- Update **Coercion** / related compatibility copy: Standard JSON Schema remains the default path; `toJsonSchema` is the documented escape hatch for converters that live outside the schema.
- Link cookbooks from core-concepts / validating-env where relevant.

**Example**

- Update `examples/with-valibot` to use ArkEnv coercion + `toJsonSchema` (add `@valibot/to-json-schema`).

### Locked (maintainer)

1. **Throw:** user `toJsonSchema` throw → immediate `ArkEnvError` for that key. Built-in probes still swallow.
2. **Skip:** `undefined` / falsy-as-undefined skip that key only; siblings still coerce.
3. **Types:** `(schema: StandardSchemaV1) => object | undefined`. No new `JsonSchema` type. No `Record<string, unknown>` (vendor closed interfaces). No `arkenv()` overloads. No `{ toJsonSchema }` shorthand.
4. **Non-plain object return:** `ArkEnvError` (same as throw). Internal `toString.call === "[object Object]"` helper; do not reuse schema-map guards.

### Acceptance criteria

- [ ] `toJsonSchema` exists on `ParseStandardConfig` / `StandardEnvConfig` as `(schema: StandardSchemaV1) => object | undefined` and enables Valibot `v.number()` / `v.boolean()` ArkEnv coercion when wired to `@valibot/to-json-schema`.
- [ ] Fallback-only: Standard JSON Schema fields coerce without the callback; the callback is not invoked for those keys.
- [ ] Hybrid Zod + Valibot schema coerces both sides when the callback is provided.
- [ ] Callback throw for one key fails the parse with `ArkEnvError` naming that key (does not swallow; does not continue as missing-JSON-Schema).
- [ ] `undefined` return skips only that key; siblings still coerce; missing-JSON-Schema hint still applies to the skipped key.
- [ ] Non-plain object return (`Date`, array, etc.) fails the parse with `ArkEnvError` for that key.
- [ ] `@arkenv/standard` has no required Valibot / `@valibot/to-json-schema` dependency. Real Valibot coverage may use test-only deps on the package that already hosts standard-mode tests.
- [ ] Tests cover: Valibot-via-callback, hybrid Zod+Valibot, throw → `ArkEnvError`, undefined skip, non-plain object → `ArkEnvError`, callback not called when probes succeed, omitted option / `coerce: false` unchanged.
- [ ] Guides → Validators ships full ArkType, Zod, and Valibot cookbooks on `simplify-docs`.
- [ ] Coercion / compatibility copy documents the escape hatch (v1 pages + Simplify Docs copies).
- [ ] `examples/with-valibot` uses ArkEnv coercion via the documented wrapper and depends on `@valibot/to-json-schema`.
- [ ] Changeset is **minor** for `@arkenv/standard` (v1 SemVer). Also bump published packages that bundle the shared standard parser if they must expose the option at runtime.

### Out of scope

- `@arkenv/standard/valibot` (or any vendor) subpath
- Separate `@arkenv/valibot` package
- Optional-peer / `vendor === "valibot"` auto-detect
- Non–JSON Schema Valibot coercion
- Per-key `toStandardJsonSchema(...)` as the documented happy path
- Adding `toJsonSchema` to ArkType `ArkEnvConfig`
- Bare `{ toJsonSchema }` function-reference shorthand
- `arkenv()` overloads / extra type params to make the vendor function assignable
- Changing `assertStandardSchemaMap` / schema-map “plain object” semantics

### Implementation notes

- Primary touch: shared standard parse / JSON Schema extraction → `@arkenv/standard`; framework plugins that reuse `ParseStandardConfig` inherit the option.
- Aligns with ADR 0002 (JSON Schema pre-coercion) and ADR 0007 (`@arkenv/standard` stays dependency-free).
- This is a **v1** issue. Do **not** implement on `dev` (different package layout; no `@arkenv/standard` split).
- Runtime, tests, example, changeset: PR into **`v1`**.
- Guides → Validators and Simplify Docs coercion/engine copy: PR into **`simplify-docs`** (tracking #1527). Do not graft a Guides section onto the pre-restructure docs tree.
- Closing the issue needs both PRs. If only one can land first, land the `v1` runtime PR first.
- Until the docs restructure merges, also correct the current v1 Coercion / `@arkenv/standard` pages so they no longer imply Valibot coercion works with no wiring step.

yamcodes:
## Summary

Implements the **runtime** half of #1564 on `v1`: optional `toJsonSchema` on `ParseStandardConfig` / `StandardEnvConfig` so Valibot (and similar) can supply JSON Schema for ArkEnv pre-coercion when Standard JSON Schema is not embedded on the schema value.

Companion docs PR (Guides → Validators on Simplify Docs): #1570.

## Changes

- Add `toJsonSchema?: (schema: StandardSchemaV1) => object | undefined` to `ParseStandardConfig`
- Call it **only as a fallback** after built-in Standard JSON Schema probes fail
- `undefined` / falsy → skip that key only; throw / non-plain object → `ArkEnvError` (`INVALID_SCHEMA`) for that key
- Internal plain-object check via `Object.prototype.toString.call === "[object Object]"`
- Tests: Valibot-via-callback, hybrid Zod+Valibot, throw, undefined skip, non-plain object, probe success / `coerce: false` / omitted option
- Update `examples/with-valibot` to use ArkEnv coercion + documented wrapper (`@valibot/to-json-schema@^1.6`); fix HOST regex for JSON Schema; add `.env.example`
- Interim v1 docs updates (Coercion / `@arkenv/standard` / Options)
- Minor changeset for `@arkenv/standard` (+ `@arkenv/core` via fixed group); size-limit 3.6 → 3.7 kB

## Test plan

- [x] `pnpm --filter @repo/utils build && pnpm --filter @arkenv/standard build`
- [x] `pnpm --filter @arkenv/standard size` (3646 / 3700)
- [x] `pnpm --filter @arkenv/core exec vitest run src/standard-mode.test.ts` (29 passed)
- [x] `pnpm run typecheck`
- [x] Local Valibot smoke with `toJsonSchema` + example schema

**Note:** `sync-examples` is already red on `v1` HEAD (unrelated playground/example `package.json` drift); this PR does not touch those synced examples (`with-valibot` is not playground-synced).

<div><a href="https://cursor.com/agents/bc-79071eb8-d629-42ff-a000-5c9f532d61dc?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-79071eb8-d629-42ff-a000-5c9f532d61dc&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -62,13 +62,11 @@
 		"@size-limit/esbuild-why": "catalog:",
 		"@size-limit/preset-small-lib": "catalog:",
 		"@types/node": "catalog:",
-		"@valibot/to-json-schema": "1.6.0",
 		"arktype": "catalog:",
 		"rimraf": "catalog:",
 		"size-limit": "catalog:",
 		"tsdown": "catalog:",
 		"typescript": "catalog:",
-		"valibot": "catalog:",
 		"vitest": "catalog:",
 		"zod": "catalog:"
 	},
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -62,7 +62,7 @@
 		"@size-limit/esbuild-why": "catalog:",
 		"@size-limit/preset-small-lib": "catalog:",
 		"@types/node": "catalog:",
-		"@valibot/to-json-schema": "1.3.0",
+		"@valibot/to-json-schema": "1.6.0",
 		"arktype": "catalog:",
 		"rimraf": "catalog:",
 		"size-limit": "catalog:",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -62,11 +62,13 @@
 		"@size-limit/esbuild-why": "catalog:",
 		"@size-limit/preset-small-lib": "catalog:",
 		"@types/node": "catalog:",
+		"@valibot/to-json-schema": "1.3.0",
 		"arktype": "catalog:",
 		"rimraf": "catalog:",
 		"size-limit": "catalog:",
 		"tsdown": "catalog:",
 		"typescript": "catalog:",
+		"valibot": "catalog:",
 		"vitest": "catalog:",
 		"zod": "catalog:"
 	},
--- a/packages/internal/utils/src/utils/standard-helpers.ts
+++ b/packages/internal/utils/src/utils/standard-helpers.ts
@@ -94,8 +94,7 @@ export function extractJsonSchema(
 			try {
 				converted = toJsonSchema(validator as StandardSchemaV1);
 			} catch (error) {
-				const detail =
-					error instanceof Error ? error.message : String(error);
+				const detail = error instanceof Error ? error.message : String(error);
 				throw new ArkEnvError([
 					buildEnvIssue(
 						key,
--- a/packages/internal/utils/src/utils/standard-helpers.ts
+++ b/packages/internal/utils/src/utils/standard-helpers.ts
@@ -15,7 +15,7 @@ function isPlainObject(value: unknown): value is Record<string, unknown> {
  * Extract JSON Schema definitions from standard schema validators.
  *
  * @param def The schema dictionary mapping keys to validators
- * @param toJsonSchema Optional fallback converter when built-in Standard JSON Schema probes fail
+ * @param toJsonSchema Optional fallback converter when a key has no Standard JSON Schema on the value
  * @returns The generated JSON Schema, a flag indicating if any JSON Schema was found,
  *          and a list of keys that do not support JSON Schema
  * @throws {ArkEnvError} When `toJsonSchema` throws or returns a non-plain object for a key
--- a/packages/internal/utils/src/utils/standard-helpers.ts
+++ b/packages/internal/utils/src/utils/standard-helpers.ts
@@ -1,5 +1,5 @@
 import type { StandardSchemaV1 } from "@repo/types";
-import { ArkEnvError } from "../core";
+import { ArkEnvError } from "@/core";
 import { buildEnvIssue } from "./errors";
 
 /**
--- a/packages/internal/utils/src/utils/standard-helpers.ts
+++ b/packages/internal/utils/src/utils/standard-helpers.ts
@@ -1,11 +1,29 @@
+import type { StandardSchemaV1 } from "@repo/types";
+import { ArkEnvError } from "../core";
+import { buildEnvIssue } from "./errors";
+
+/**
+ * Whether `value` is a plain object (`{}` / Object.create(null) style).
+ * Rejects arrays, `Date`, functions, boxed primitives, etc.
+ * @internal
+ */
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+	return Object.prototype.toString.call(value) === "[object Object]";
+}
+
 /**
  * Extract JSON Schema definitions from standard schema validators.
  *
  * @param def The schema dictionary mapping keys to validators
+ * @param toJsonSchema Optional fallback converter when built-in Standard JSON Schema probes fail
  * @returns The generated JSON Schema, a flag indicating if any JSON Schema was found,
  *          and a list of keys that do not support JSON Schema
+ * @throws {ArkEnvError} When `toJsonSchema` throws or returns a non-plain object for a key
  */
-export function extractJsonSchema(def: Record<string, unknown>): {
+export function extractJsonSchema(
+	def: Record<string, unknown>,
+	toJsonSchema?: (schema: StandardSchemaV1) => object | undefined,
+): {
 	jsonSchema: Record<string, any>;
 	hasJsonSchema: boolean;
 	missingKeys: string[];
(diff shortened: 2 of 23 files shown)


packages/standard/src/standard-mode.test.ts
yamcodes, line 1: why is this test in `/core` ?  isnt it standard specific ?

packages/core/package.json
yamcodes, line 1: why we need this change in `/core` ? shouldnt this be part of `/standard`?

packages/internal/utils/src/utils/standard-helpers.ts
```diff
@@ -1,11 +1,29 @@
+import type { StandardSchemaV1 } from "@repo/types";
+import { ArkEnvError } from "../core";
```
yamcodes: prefer importing from `@/*` than from `../*`

PR #1570, opened 2026-08-12
feat(standard): toJsonSchema escape hatch + Validators guides (#1564)

### LINKED ISSUE #1564: (v1) options.toJsonSchema: Valibot-friendly coercion escape hatch + Validators guides

yamcodes:
## Summary

#1564 on `simplify-docs`: optional `toJsonSchema` coercion escape hatch for `@arkenv/standard`, plus **Guides → Validators** cookbooks (ArkType, Zod, Valibot).

Merge into `simplify-docs`, then into `v1` when ready.

Typing decision: **always-`as`** at host-converter calls. Callback is `StandardSchemaV1`. Docs/examples assert `as v.GenericSchema` / `as z.ZodMiniType` even for Valibot-only / Mini-only. Evaluation notes: `docs/to-json-schema-wiring.md`, deck: `docs/to-json-schema-s-tier-deck.md`.

## Runtime

- `toJsonSchema?` on `ParseStandardConfig` / `StandardEnvConfig` as `(schema: StandardSchemaV1) => object | undefined`
- Single `arkenv` signature so the callback is contextually typed (`safe: true` still returns `SafeArkEnvResult`)
- Per-key fallback in `extractJsonSchema` when a value has no Standard JSON Schema
- `undefined` skips that key; throw / non-plain object → `ArkEnvError` (`INVALID_SCHEMA`)
- Tests under `@arkenv/standard`; Valibot example uses the documented wrapper with `as v.GenericSchema`
- Changeset included (minor for `@arkenv/standard`)

## Docs

- New Guides → Validators section (index + ArkType + Zod + Valibot)
- Zod cookbook: Standard JSON Schema on the value
- Valibot cookbook: documented wrapper with `as v.GenericSchema` (`typeMode: "input"`, `target: "draft-07"`) — **first** `toJsonSchema` example
- Zod Mini: off-value converter via `z.toJSONSchema`; mixing uses vendor switch + assertions
- Coercion / Choosing an engine / Options / `@arkenv/standard` updates

## Test plan

- [x] `pnpm --filter @arkenv/standard test:once`
- [x] Twoslash on Valibot + Coercion examples
- [x] Twoslash on Valibot + Zod Mini mixing example
- [x] Twoslash on Valibot-only and Zod + Valibot with `GenericSchema` assertion
- [x] `pnpm --filter www exec vitest run lib/twoslash-options.test.ts`
- [ ] Preview docs nav: Guides → Validators → ArkType / Zod / Valibot
- [ ] Spot-check links from Coercion and Choosing an engine


yamcodes:
LGTM

PR #1580, opened 2026-08-15
fix: (v1) pnpm www exits with "Another next dev server is already running"

### LINKED ISSUE #1578: (v1) pnpm www exits with "Another next dev server is already running"
## Problem

`pnpm www` (Turbo `www#dev`) prints **Ready**, then exits 1 with:

```
Another next dev server is already running.
- Local:        http://localhost:3000
- PID:          <pid>
- Dir:          apps/www
Run kill <pid> to stop it.
```

The docs site is usually still serving on port 3000 as an **orphaned** `next-server`. Killing that PID and rerunning reproduces the same failure with a **new** PID — this is not a leftover process from a previous session.

## Current behavior

The `www` `dev` script starts two processes at once via `concurrently`:

- `next dev`
- `next-video sync -w`

`next-video` loads `next.config` as `phase-development-server` to read video config. That re-evaluates fumadocs `createMDX()`, which writes `.source` files **after** Next has already printed Ready.

Next.js 16.2 then treats the reload as a second `next dev` in the same directory (`.next/dev/lock`), the CLI exits 1, and Turbo reports `www#dev` failed. `next-video sync -w` also exits 0 instead of staying in watch mode, so `concurrently` (success = all) tears the script down.

`kill <pid>` is the wrong recovery: each `pnpm www` run races itself.

## Desired behavior

`pnpm www` stays in the foreground as a single healthy `next dev`. Video assets can still be processed, but **not** by a second process that loads Next's config while the dev server is starting.

**Approach:** drop `next-video sync -w` from the concurrent `dev` script. Keep `next dev` as the only long-lived process. If video sync is still needed, run a one-shot `next-video sync` (no `-w`) in `predev`, where it cannot race the lock.

## Acceptance criteria

- [ ] `pnpm www` (or `pnpm --filter www dev`) stays running; it does not exit 1 with the Next.js 16.2 lock error during a clean start.
- [ ] http://localhost:3000 serves the docs site from that same process (not an orphan from a previous failed start).
- [ ] Stopping the command (Ctrl+C) actually stops the server on port 3000.
- [ ] Adding or updating a docs video still has a documented/maintainer path (one-shot sync in `predev` is enough; live `videos/` watch during `dev` is not required).

## Out of scope

- Changing Next.js, fumadocs-mdx, or next-video themselves.
- Docs content, playgrounds, or examples.
- Keeping `next-video sync -w` beside `next dev` (that coupling is the bug).

## Repro

1. From the repo root, ensure nothing is listening on port 3000.
2. Run `pnpm www`.
3. Observe Ready, then `next-video` MDX generation, then the lock error and a non-zero exit — while port 3000 may still be held by an orphan `next-server`.

yamcodes:
Fixes #1578

Stop racing Next.js 16.2's `.next/dev/lock` by dropping the parallel `next-video sync -w` process from the www `dev` script. `pnpm www` now stays in the foreground as a single `next dev`, with a one-shot `next-video sync` in `predev` so docs videos can still be processed without a live watcher.

## Test plan
- [ ] From a clean checkout with nothing on port 3000, run `pnpm www` and confirm it stays running (no "Another next dev server is already running")
- [ ] Confirm http://localhost:3000 is served by that same process
- [ ] Ctrl+C the command and confirm port 3000 is freed
- [ ] After adding or replacing a file in `apps/www/videos/`, restart `pnpm www` (or run `pnpm exec next-video sync` from `apps/www`)


yamcodes:
Addressed the stale VS Code task detail in `08ba2ae`. `apps/www/.vscode/tasks.json` now shows `next dev`, matching the current `dev` script.

PR #1702, opened 2026-08-29
feat: Harden schema capture: fail closed, distinct inspect errors, document env.ts contract

### LINKED ISSUE #1636: Harden schema capture: fail closed, distinct inspect errors, document env.ts contract
## Summary

Keep schema capture as the inspect primitive for `arkenv sync` / `arkenv check` (ADR 0027, loader in #1622). Do not switch to `{ safe: true }`, a public skip-validate flag, source copy/inject, or a required `export const schema` split.

Harden the loader so inspect **fails closed** when it cannot honestly report keys, split error codes that today share one “upgrade core” hint, and document what `env.ts` may and must not do under capture.

Blocked on / related: #1234 (`sync`), #962 (`check`), #1314 (import, don’t parse).

## Status quo

- Jiti-import the user’s flat schema module with `beginSchemaCapture()` / `endSchemaCapture()` (unpublished `globalThis` handshake). User’s installed `@arkenv/core` / `@arkenv/standard` must honor the flag.
- `arkenv()` records the def and returns `{}`.
- `NO_SCHEMA` (zero captured calls) and an old runtime that still validates both look like “upgrade core.”
- A def that yields no keys can look like a valid `arkenv({})`.
- `hasDefault` is best-effort (ArkType / Zod / Valibot). Unknown Standard Schema still lists keys.
- Loader is one `schemaPath` (flat). Strict `env/server.ts` + `env/client.ts` is not merged.
- Nuxt already uses the same `globalThis` key with a different bag shape.

## Desired behavior

1. **Fail closed on dishonest empty schemas.** `arkenv({})` → success, `keys: []`. Zero `arkenv()` calls, or a captured value we cannot read keys from, → not success. Never let `sync` treat that as “empty schema, rewrite `.env.example`.”
2. **Distinct, stable inspect error codes** (names can be bikeshed at implementation), at least:
   - no `arkenv()` call at module load
   - installed core/standard too old (validation ran instead of capture)
   - captured def not extractable
   - module threw because it **used** `env` values at load (existing contract hint)
3. **`hasDefault` is comments-only.** `sync` must still emit the key when `hasDefault` is false (BYO / Standard Schema). Do not omit keys because the sniffer missed a default.
4. **Document the inspect contract** (CONTEXT + CLI/schema-loader docs, not only JSDoc):
   - **Supported:** top-level `export const env = arkenv({ ... })` (ArkType strings, compiled `type({...})`, Standard Schema map including unknown vendors); spreads from other modules; `arkenv({})`; `env.NODE_ENV === "production"` (stub is `{}`).
   - **Unsupported (inspect refuses or is incomplete):** using `env` values at module scope (`createClient(env.X)`, `if (!env.X) throw`); `{ safe: true }` then reading `env.success`; lazy `export const env = () => arkenv(...)`; `arkenv()` behind a runtime branch; old core; import-time side effects; pointing the loader at only one strict-layout file.
5. **Handshake:** keep unpublished skip-parse. Do **not** add a public `schemaMode` / `dryRun` on `ArkEnvConfig` (production off-switch + autocomplete). Optional hardening: version/namespace the bag so it does not collide with Nuxt’s capture state; prefer `INSPECT_UNSUPPORTED` over `NO_SCHEMA` when the flag is ignored.

## Out of scope

- Required `export const schema` + `export const env = arkenv(schema)` API
- `{ safe: true }` as the inspect loader
- Copying `env.ts` and injecting a flag
- Jiti-aliasing the user’s core to the CLI’s runtime
- Publishing `beginSchemaCapture` on the app barrel
- Implementing `sync` / `check` themselves (#1234, #962)
- Implementing strict-layout multi-file loading (document as incomplete; ship with those commands if needed)
- Making `hasDefault` complete for every Standard Schema vendor

yamcodes:
Fixes #1636

## Summary
- Fail closed when schema capture cannot honestly report keys (`arkenv({})` stays success with `keys: []`)
- Split loader failures into locked codes: `ERR_INSPECT_NO_CALL`, `ERR_INSPECT_UNSUPPORTED`, `ERR_INSPECT_UNEXTRACTABLE`, `ERR_INSPECT_EVAL_THROW`
- Move CLI/core/standard capture handshake to `Symbol.for("arkenv.schemaCapture.v1")` (Nuxt stays on `__ARKENV_SCHEMA_CAPTURE__`)
- Document the inspect `env.ts` contract (hollow `{}` stub, not a Proxy) in CONTEXT, ADR 0027, and `example` / `check` docs

## Test plan
- [x] Loader: `arkenv({})` → `ok: true`, `keys: []`
- [x] Loader: no `arkenv()` call → `ERR_INSPECT_NO_CALL`
- [x] Loader: simulated old runtime → `ERR_INSPECT_UNSUPPORTED`
- [x] Loader: non-extractable def → `ERR_INSPECT_UNEXTRACTABLE`
- [x] Loader: env use at module scope → `ERR_INSPECT_EVAL_THROW`
- [x] Core: Symbol handshake isolated from Nuxt legacy string key
- [x] `pnpm typecheck`
- [x] Package tests for `arkenv` / `@arkenv/core` / `@arkenv/standard` (pre-existing scaffold/hero Zod assertion failures on `v1` are unrelated)

### DIFF
--- a/.changeset/harden-schema-capture-inspect.md
+++ b/.changeset/harden-schema-capture-inspect.md
@@ -16,4 +16,4 @@ The CLI schema loader now fails closed when it cannot honestly report keys, and 
 
 `arkenv({})` remains a valid empty schema. Capture state for core, standard, and the CLI now lives on `Symbol.for("arkenv.schemaCapture.v1")` so it does not collide with Nuxt’s legacy string key. Upgrade `@arkenv/core` / `@arkenv/standard` alongside the CLI so inspect can see the capture flag.
 
-MDX tables in `@arkenv/fumadocs-ui` are now keyboard-focusable so wide comparison tables pass `scrollable-region-focusable`.
+MDX table wrappers in `@arkenv/fumadocs-ui` now include `fd-scroll-container` so wide comparison tables are excluded from `scrollable-region-focusable` the same way as code blocks.
--- a/.changeset/harden-schema-capture-inspect.md
+++ b/.changeset/harden-schema-capture-inspect.md
@@ -2,6 +2,7 @@
 "arkenv": minor
 "@arkenv/core": patch
 "@arkenv/standard": patch
+"@arkenv/fumadocs-ui": patch
 ---
 
 #### Harden schema capture for fail-closed inspect
@@ -14,3 +15,5 @@ The CLI schema loader now fails closed when it cannot honestly report keys, and 
 - `ERR_INSPECT_EVAL_THROW` — the module threw while using env values at load time
 
 `arkenv({})` remains a valid empty schema. Capture state for core, standard, and the CLI now lives on `Symbol.for("arkenv.schemaCapture.v1")` so it does not collide with Nuxt’s legacy string key. Upgrade `@arkenv/core` / `@arkenv/standard` alongside the CLI so inspect can see the capture flag.
+
+MDX tables in `@arkenv/fumadocs-ui` are now keyboard-focusable so wide comparison tables pass `scrollable-region-focusable`.
--- a/.changeset/harden-schema-capture-inspect.md
+++ b/.changeset/harden-schema-capture-inspect.md
@@ -1,0 +1,16 @@
+---
+"arkenv": minor
+"@arkenv/core": patch
+"@arkenv/standard": patch
+---
+
+#### Harden schema capture for fail-closed inspect
+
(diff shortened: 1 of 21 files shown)


yamcodes:
Addressed the fumadocs-ui release-hygiene note: moved the `fd-scroll-container` table wrap into `apps/www/mdx-components.tsx` and reverted `@arkenv/fumadocs-ui`, so the published package is unchanged and needs no changeset bump.

